PR #15, opened 2023-02-28
:seedling: Allow to customize the ISO for airgap

mudler:
This changeset introduces few changes:

- allow to setup additional data to embed into the iso with `--set iso.data=<path>`
- allows to setup a netboot cmdline with `--set netboot.cmdline=<>`
- Various optimizations
- A full airgap e2e example to bundle k3s images

Fixes: https://github.com/kairos-io/kairos/issues/976


mudler:
Docs will be a separate PRs on kairos referencing the example here

mudler:
Testing this locally and works as expected, merging

PR #16, opened 2023-03-14
Add release job for binaries

Itxaka:
This provides binaries for all systems so its easier to get them and run in both macos and windows


mudler (approved):
Looks good!

PR #18, opened 2023-03-17
:seedling: Generate disks

mudler:
This adds disks generation to AuroraBoot.

Still WIP as I'd like to add e2e tests at least.


mudler:
I refrained to do anything there as we will move the build code out of it. Scripts are doing different things than elemental sadly. I will expect to have all the building bits here instead on the long run

.github/workflows/tests.yml
```diff
@@ -0,0 +1,20 @@
...
+  tests:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Run tests
+        run: |
+          docker run --privileged -v /var/run/docker.sock:/var/run/docker.sock --rm -t -v $(pwd):/workspace -v earthly-tmp:/tmp/earthly:rw earthly/earthly:v0.6.21 --allow-privileged +test
```
Itxaka, line 20: why 0.6.21?

Earthfile
```diff
@@ -0,0 +1,16 @@
...
+test:
+    FROM alpine
+    WORKDIR /test
+    RUN apk add go docker jq
+    ENV GOPATH=/go
+    COPY . .
+    WITH DOCKER \
+            --allow-privileged \
+            --load auroraboot:latest=+image
+        RUN go install -mod=mod github.com/onsi/ginkgo/v2/ginkgo && /go/bin/ginkgo -r -p --randomize-all --procs 2 --fail-fast --flake-attempts 3 ./...
```
Itxaka, line 15: if ginkgo is in teh go.mod file you can just use `go run` and it will pick the proper version!

Itxaka (approved):
Looking good!

Earthfile
mudler: tried that, but didn't worked as we don't have it as we don't include ginkgo in the code, only in the tests :/

.github/workflows/tests.yml
mudler: just copy-paste, I can update it

mudler: let me do that in a follow up, don't want to break tests now that are green :sweat_smile:

PR #34, opened 2023-07-05
Disable squashfs compression on iso creation

Itxaka:
We found some time ago that the compression is quite slow and it doesnt bring a big improvement on final image size.

Fixes https://github.com/kairos-io/kairos/issues/1553


mauromorales (approved):
lgtm

PR #35, opened 2023-08-01
Dont panic on template parse failure and change delims

Itxaka:
the default delimitators were changed to be [[ ]] but unfortunately there is a high chance of hitting those delimitators in a cloud config file that can be using commands in its steps.

This patch changes those delimitators to [[[ ]]] which is much more difficult to hit.

It also reworks the render method to be more careful to avoid raising a panic and leaving the user with no clue on what happened. Now it properly checks for errors when parsing the data and returns that.

It alos logs the error properly to the logger and to debug with much more data so its easier to troubleshoot

Reported by Dave Sargent on Slack.

### DIFF
--- a/internal/cmd/config.go
+++ b/internal/cmd/config.go
@@ -19,6 +19,9 @@ import (
 	"gopkg.in/yaml.v1"
 )
 
+const delimLeft = "[[["
+const delimright = "]]]"
+
 func isUrl(s string) bool {
 	u, err := url.Parse(s)
 	return err == nil && u.Scheme != "" && u.Host != ""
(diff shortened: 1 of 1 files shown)


internal/cmd/config.go
```diff
@@ -48,12 +51,19 @@ func downloadFile(url string) (content string, err error) {
 	return b.String(), nil
 }
 
-func render(data string, foo any) string {
-	t := template.New("cloudConfig template").Delims("[[", "]]").Option("missingkey=zero")
-	t, _ = t.Parse(data)
```
jimmykarily, line 53: :man_facepalming: we should stop hiding errors... Good catch!

jimmykarily (approved):
I can't find any reference of the old delimiters in docs, was it only used internally in our code somehow? In any case, this PR looks good. Kudos!

PR #36, opened 2023-08-09
Disable selinux by default in ipxe boot

Itxaka:
ipxe is like a livecd, so we need to disable selinux to avoid running into permissions issues while ipxe booting

### DIFF
--- a/pkg/ops/netboot.go
+++ b/pkg/ops/netboot.go
@@ -29,7 +29,7 @@ func StartPixiecore(cloudConfigFile, squashFSfile, address, netbootPort, initrdF
 
 		configFile := cloudConfigFile
 
-		cmdLine := `rd.live.overlay.overlayfs rd.neednet=1 ip=dhcp rd.cos.disable root=live:{{ ID "%s" }} netboot nodepair.enable config_url={{ ID "%s" }} console=tty1 console=ttyS0 console=tty0`
+		cmdLine := `rd.live.overlay.overlayfs selinux=0 rd.neednet=1 ip=dhcp rd.cos.disable root=live:{{ ID "%s" }} netboot nodepair.enable config_url={{ ID "%s" }} console=tty1 console=ttyS0 console=tty0`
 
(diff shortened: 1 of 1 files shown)


pkg/ops/netboot.go
```diff
@@ -29,7 +29,7 @@ func StartPixiecore(cloudConfigFile, squashFSfile, address, netbootPort, initrdF
 
 		configFile := cloudConfigFile
 
-		cmdLine := `rd.live.overlay.overlayfs rd.neednet=1 ip=dhcp rd.cos.disable root=live:{{ ID "%s" }} netboot nodepair.enable config_url={{ ID "%s" }} console=tty1 console=ttyS0 console=tty0`
+		cmdLine := `rd.live.overlay.overlayfs selinux=0 rd.neednet=1 ip=dhcp rd.cos.disable root=live:{{ ID "%s" }} netboot nodepair.enable config_url={{ ID "%s" }} console=tty1 console=ttyS0 console=tty0`
```
mauromorales, line 32: Can you test the Ubuntu image with this change? I remember there being issues trying to start it without selinux enabled

Itxaka: any specific version? latest?

mauromorales: yeah latest as long as it's the `ubuntu` only flavor, not lts

Itxaka: will test, thanks for the tip!

Itxaka: @mauromorales used latest core-ubuntu:v2.3.1 artifact and was able to boot and install with no issues!

mauromorales: fantastic!

mauromorales (approved):

Itxaka:
@mudler any idea why this could be failing ??? By changing one string???? In UNRELATED TESTS ????????????

Itxaka:
This will be reworked to be set in the base image instead

PR #45, opened 2023-10-04
:warning: add mbr disk generation

mudler:
supersedes https://github.com/kairos-io/osbuilder/pull/98

- Adds MBR disk generation
- Adds e2e tests for MBR disk generation
- Switch tests to use self-hosted runners
- Do not use earthly anymore in the CI runs
- Run tests with KVM when necessary
- Fixes https://github.com/kairos-io/kairos/issues/1894
- Output artifacts now are moved in the `build` directory instead of `iso`. The name was changed since we build also disks and other assets and it is more generic.


deployer/register.go
```diff
@@ -142,15 +144,28 @@ func Register(g *herd.Graph, artifact schema.ReleaseArtifact, c schema.Config, c
...
+
+	g.Add(opGenMBRRawDisk,
+		herd.EnableIf(func() bool { return c.Disk.ARM == nil && c.Disk.MBR }),
+		herd.IfElse(isoOption(),
+			herd.WithDeps(opDownloadISO), herd.WithDeps(opGenISO),
+		),
+		herd.IfElse(isoOption(),
+			herd.WithCallback(
+				ops.GenBIOSRawDisk(isoFile, filepath.Join(dst, "disk.raw"), true)),
+			herd.WithCallback(
```
mudler, line 165: this is not right yet

mudler:
All fixed!

pkg/ops/disks.go
```diff
@@ -117,7 +117,68 @@ func GenArmDisk(src, dst string, do schema.Config) func(ctx context.Context) err
...
+        -rtc base=utc,clock=rt \
+        -chardev socket,path=qga.sock,server,nowait,id=qga0 \
+        -device virtio-serial \
+        -device virtserialport,chardev=qga0,name=org.qemu.guest_agent.0 \
+        -drive if=virtio,media=disk,file=%s \
+        -drive format=raw,media=cdrom,readonly=on,file=%s \
+        -drive format=raw,media=cdrom,readonly=on,file=ci.iso \
+        -boot d %s
+        
+`, cloudConfigFile, dst, qemuBin, ram, cores, dst, srcISO, extra),
```
mauromorales, line 171: I would prefer if these commands were separated by different Sprintf calls, each with their own values, just because it will make it easier to map that the `cloudConfigFile` is for the `cp`, the `dst` is for `mkisofs` and so on .... but not a blocker, just a personal view

deployer/register.go
```diff
@@ -142,15 +144,28 @@ func Register(g *herd.Graph, artifact schema.ReleaseArtifact, c schema.Config, c
...
 
 	g.Add(opGenRawDisk,
-		herd.EnableIf(func() bool { return diskIsSet && c.Disk.ARM == nil }),
+		herd.EnableIf(func() bool { return rawDiskIsSet && c.Disk.ARM == nil && !c.Disk.MBR }),
 		imageOrSquashFS,
-		herd.WithCallback(ops.GenRawDisk(tmpRootfs, filepath.Join(dst, "disk.raw"))))
+		herd.WithCallback(ops.GenEFIRawDisk(tmpRootfs, filepath.Join(dst, "disk.raw"))))
+
+	g.Add(opGenMBRRawDisk,
+		herd.EnableIf(func() bool { return c.Disk.ARM == nil && c.Disk.MBR }),
```
mauromorales, line 158: why don't we need to validate here `rawDiskIsSet`?

Itxaka:
Not sure if I understand this correctly, we are using qemu to generate a disk???

we already had disk creation on elemental in pure go (minus copying the fat files needed, that required a shell call) before? Why go over to this? This seems way too much IMHO. Another dep to add to aurora and probably not too small either and behaving differently for sure between versions and systems...


Im not sure I like this more than https://github.com/kairos-io/elemental-cli/blob/main/pkg/action/build-disk.go#L40

At this point, why not use terraform which is specialized in doing exactly this? Getting an input iso and generating a raw output image?

mudler:
> Not sure if I understand this correctly, we are using qemu to generate a disk???

Yes, pretty simple actually :)

> 
> we already had disk creation on elemental in pure go (minus copying the fat files needed, that required a shell call) before? Why go over to this? This seems way too much IMHO. Another dep to add to aurora and probably not too small either and behaving differently for sure between versions and systems...
> 
> Im not sure I like this more than https://github.com/kairos-io/elemental-cli/blob/main/pkg/action/build-disk.go#L40

What we had is only for EFI and it's static and note that we have already an equivalent mechanism in AuroraBoot. The previous approach doesn't install a system, it merely sets up a disk such as the first boot handles the deployment. In this case we actually snapshot the disk.

> 
> At this point, why not use terraform which is specialized in doing exactly this? Getting an input iso and generating a raw output image?

Why should we use terraform to generate a simple, raw disk image? Terraform makes sense IMHO to bootstrap cloud environment with sophisticated services setup -  I think a QEMU call here is more than enough

deployer/register.go
mudler: raw disks are images which are generated statically, without the need of an hypervisor - here instead we run QEMU to generate an image

pkg/ops/disks.go
mudler: gotcha, good point :+1: will try to refactor then

Itxaka:
> > Not sure if I understand this correctly, we are using qemu to generate a disk???
> 
> Yes, pretty simple actually :)
> 
> > we already had disk creation on elemental in pure go (minus copying the fat files needed, that required a shell call) before? Why go over to this? This seems way too much IMHO. Another dep to add to aurora and probably not too small either and behaving differently for sure between versions and systems...
> > Im not sure I like this more than [kairos-io/elemental-cli@`main`/pkg/action/build-disk.go#L40](https://github.com/kairos-io/elemental-cli/blob/main/pkg/action/build-disk.go?rgh-link-date=2023-10-17T14%3A40%3A34Z#L40)
> 
> What we had is only for EFI and it's static and note that we have already an equivalent mechanism in AuroraBoot. The previous approach doesn't install a system, it merely sets up a disk such as the first boot handles the deployment. In this case we actually snapshot the disk.
> 
> > At this point, why not use terraform which is specialized in doing exactly this? Getting an input iso and generating a raw output image?
> 
> Why should we use terraform to generate a simple, raw disk image? Terraform makes sense IMHO to bootstrap cloud environment with sophisticated services setup - I think a QEMU call here is more than enough

I dont know, Im conflicted... seems strigh enough but Im fearful of the qemu issues.

How about fixing kairos-agent to be able to install to a mounted loop device or something like that? Wouldnt that cover this feature and especially make it so the path is the same. which is one of the big disadvantages of the current implementation?

Alos this is only valid for x68. No arm, no rpi specific bits nor anything else right?

I dont know. It feels kind of wrong xD

mudler:
ok merging this for now as discussed in the call (cc @Itxaka @mauromorales ) - let's also keep in mind that this is mostly user-facing, we can look at the implementation bits later anytime and optimize it as needed.  There are of course many things that aren't perfect yet, but would like to go ahead as it was already challenging due to the test being broken in the way of it

PR #60, opened 2024-07-11
Bump to go1.22

Itxaka:


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/kairos-io/AuroraBoot
 
-go 1.19
+go 1.22
```
mauromorales, line 3: does this mean latest in 1.22? then we should do this for all of them

mauromorales (approved):

PR #66, opened 2024-10-02
Fail out if cc was specified but contents are empty

Itxaka:
If we have marked a cloud config file but after parsing and templating,
the contents are empty, error out as it points out to a user error on
choosing the file.

If no cc is set, the check wont fail out.

This also brings some small improvements to dockerfile to not have to fully rebuild the dockerfile each time something is changed on aurora for testing

Probably fixes https://github.com/kairos-io/kairos/issues/2876 as it looks like a wrong naming issue can lead to confusion like this. With this patch, the error will be immediately evident

Signed-off-by: Itxaka <itxaka@kairos.io>

### DIFF
--- a/tests/e2e/iso_test.go
+++ b/tests/e2e/iso_test.go
@@ -19,7 +19,7 @@ var _ = Describe("ISO image generation", Label("iso"), func() {
 
 			tempDir = t
 
-			err = WriteConfig("", t)
+			err = WriteConfig("test", t)
 			Expect(err).ToNot(HaveOccurred())
 		})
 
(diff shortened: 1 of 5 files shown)


tests/e2e/iso_test.go
```diff
@@ -44,6 +44,17 @@ var _ = Describe("ISO image generation", Label("iso"), func() {
...
+		It("fails if cloud config is empty", func() {
+			err := WriteConfig("", tempDir)
+			Expect(err).ToNot(HaveOccurred())
+
+			out, err := RunAurora(fmt.Sprintf(`--set container_image=quay.io/kairos/core-rockylinux:latest \
+			--set "disable_http_server=true" \
+			--set "disable_netboot=true" \
+			--cloud-config /config.yaml \
+			--set "state_dir=/tmp/auroraboot"`), tempDir)
+			Expect(err).To(HaveOccurred(), out)
```
jimmykarily, line 56: There is a `MatchError` matcher we can use, to make sure we are getting the expected error and not another one.
```suggestion
			Expect(err).To(MatchError("cloud config set but contents are empty"), out)
```

jimmykarily (approved):

tests/e2e/iso_test.go
Itxaka: I thought catching exact error messages was frowned upon, as the text can change over time?

Itxaka: it better still work now :P

jimmykarily (approved):

PR #68, opened 2024-10-21
1633 use enki as lib for build iso

jimmykarily:


pkg/ops/iso.go
```diff
@@ -31,10 +38,42 @@ func GenISO(name, src, dst string, i schema.ISO) func(ctx context.Context) error
 		}
 
 		log.Info().Msgf("Generating iso '%s' from '%s' to '%s'", name, src, dst)
-		out, err := utils.SH(fmt.Sprintf("/entrypoint.sh --debug --name %s build-iso --squash-no-compression --overlay-iso %s --date=false --output %s dir:%s", name, overlay, dst, src))
-		log.Printf("Output '%s'", out)
+		cfg, err := enkiconfig.ReadConfigBuild("/config", &pflag.FlagSet{})
```
jimmykarily, line 41: This code here seems to only load [this config](https://github.com/kairos-io/osbuilder/blob/405eda716a6c291eb539a98765a291e893d0eda1/tools-image/config.yaml). Given we hardcode the same setting [some lines below](https://github.com/kairos-io/AuroraBoot/pull/68/files#diff-8b2bb3456340e31af34769338d6fcf82e6eb8cbf9455501b35255138c4965943R53), can we skip this loading of external configuration?

We have a command with command line arguments. That should be enough to get user's input. I don't think we should accept configuration from all over, including hardcoded locations like this one.

pkg/ops/iso.go
```diff
@@ -31,10 +38,42 @@ func GenISO(name, src, dst string, i schema.ISO) func(ctx context.Context) error
...
+		if err != nil {
+			return err
+		}
+		cfg.Name = name
+		cfg.OutDir = dst
+		cfg.Logger = sdkTypes.NewKairosLogger("enki", "debug", false)
+		imgSource, err := v1.NewSrcFromURI("dir:" + src)
```
Itxaka, line 48: why not `v1.NewDirSrc(src)` ? we expect the source to be a valid dir already no? If it needs validation it would be on the CLI side, not here?

pkg/ops/iso.go
Itxaka: yes please, its time to clean this up

Itxaka (approved):

pkg/ops/iso.go
jimmykarily: Done, changed to NewDirSrc

pkg/ops/iso.go
```diff
@@ -31,10 +38,38 @@ func GenISO(name, src, dst string, i schema.ISO) func(ctx context.Context) error
...
+
+		// Removed here: https://github.com/kairos-io/osbuilder/commit/866dc42c48878fa8bebab26d8f24bf0f2953eb6e#diff-5e18c6520d9fb094d1d4c52b4fb652f09becd65053d95f8d7ac89c7998172fd4
+		// enki is called here: https://github.com/kairos-io/AuroraBoot/blob/9ca30fc6e3b6cac2d227e7937bd6895198bdc4d0/pkg/ops/iso.go#L34
+		// config is loaded here: https://github.com/kairos-io/enki/blob/844bd15261e89f103c16d1caa656ec41524b8d4f/pkg/config/config.go#L110-L114
+		// config was copied in place here: https://github.com/kairos-io/osbuilder/blob/405eda716a6c291eb539a98765a291e893d0eda1/tools-image/Dockerfile#L86
+		// which is part of the auroraboot image: https://github.com/kairos-io/AuroraBoot/blob/9ca30fc6e3b6cac2d227e7937bd6895198bdc4d0/Dockerfile#L1
+		// TODO: Cleanup this madness!
+
+		//uefi := v1.NewDirSrc("/efi")
+		//spec.UEFI = append(spec.UEFI, uefi)
```
jimmykarily, line 60: `/efi` and `/grub2` were populated here: https://github.com/kairos-io/osbuilder/blob/95509370f6a87229879f1a381afa5d47225ce12d/tools-image/Dockerfile#L29-L30

I skipped `/efi` and it still builds a bootable iso (in qemu at least) but it's not the case for `/grub2`. It's still needed. Are we sure the contents of `/efi` are now found in the rootfs image? @Itxaka any link on where this happens?

There are just 2 luet packages and luet is a golang binary. Would it be better if I just installed those 2 packages here instead of relying on the Dockerfile of the tools image to do that?

A better approach is to make the arguments of this command and allow the caller to set those directories. After all, it's the caller that knows where they installed those luet packages. And if those directories are required, we make them required arguments.

pkg/ops/iso.go
jimmykarily: done

pkg/ops/iso.go
Itxaka: copyShim and copyGrub both search in the rootfs for the files I think:

https://github.com/kairos-io/enki/blob/main/pkg/action/build-iso.go#L256
```
// copyShim copies the shim files into the EFI partition
// tempdir is the temp dir where the EFI image is generated from
// rootdir is the rootfs where the shim files are searched for
```
The only part in which we would fail is on alpine which doesnt ship signed efi files so we use our packaged ones:

```
		// All failed...maybe we are on alpine which doesnt provide shim/grub.efi ?
		// In that case, we can just use the luet packaged artifacts
```

In which case, we could provably just default to the ones they ship, after all alpine wont work with secureboot due to this, so we migth as well ship their own provided files and drop the luet packages once and for all.

Let me check the grub2 stuff

Itxaka: yeah grub seems to do the same things, gets a list of possible grub2 efi files, then searches the rootfs for them to copy them into the EFI image for the iso... maybe the rootfs its searching it for its wrong?

jimmykarily: For now, I added them both back and a comment to indicate where they are coming from. Let's merge it as it is and clean up more in a future iteration.

Itxaka: I think rootfs sh ould provde all those files, we cannot go backwards and let the user provide those, we already went that route and it was a mess, secureboot broken, different paths for everything and such. Distros already provide all the needed files and in the case of needing a generic file like eltorito.img, we could even ship a generic one bundled with auroraboot as its generic enough to work for everything.

jimmykarily: `copyEfi` and `copyGrub` both copy files from `/efi` . Nothing copies files [from `/grub2`](https://github.com/kairos-io/osbuilder/blob/95509370f6a87229879f1a381afa5d47225ce12d/tools-image/Dockerfile#L29).
I guess I can remove `/efi` at least from here.

mauromorales (approved):

Itxaka:
I had a quick look, the /grub2 location provides grub BIOS files for arm and x86 hybrid boot.

hybrid boot means we build isos that can boot both on BIOS and EFI.

for this we need syslinux and grub2 files for x86_64, arm64 AND i386

arm64 does not have BIOS boot but enki iso builder does not do any distinction and when building isos it tries to build hybrid issues no matter the arch.

i386 files are needed to build the eltorito.img which is used for the hybrid boot


 Things that we can do down the road
 - make isobuilder aware that for arm64 we dont build hybrid isos, so the call to xorriso needs to be different (this means no grub2 arm64 packages needed)
 - try to get the grub2 artifacts from the rootfs (this mneans no grub2-x86_64 packages needed, they are already shipped in the rootfs)
 - move the livecd configs into the iso builder directly and allow overriding them when building.


The only problem with this is that we still need the eltorito.img and the boot_hybrid.img files in order to build hybrid isos. Not all distros provide those, nor all distros allows you to install grub2-i386 and generate one yourself!

to build a eltorito.img:

 - install grub2-i386
 - generate a grub core.img with 386 platform and some modules bundled:
```
grub2-mkimage -O i386-pc -o core.img -p /boot/grub2 -d /usr/share/grub2/i386-pc ext2 iso9660 linux echo configfile search_label search_fs_file search search_fs_uuid ls normal gzio png fat gettext font minicmd gfxterm gfxmenu all_video xfs btrfs lvm luks gcry_rijndael gcry_sha256 gcry_sha512 crypto cryptodisk test true loadenv part_gpt part_msdos biosdisk vga vbe chain boot
```
 - merge the resulting core.img with the cdboot.img(shipped by grub2-i386) `cat cdboot.img core.img > eltorito.img`

the when building the iso you pass the following to xorisso for the BIOS part:

```
-boot_image grub bin_path=eltorito.img -boot_image grub grub2_mbr=/boot_hybrid.img
```

So maybe there is an alternative approach to this with xorriso as well? Maybe we only need the eltorito.img and we can ship a prebuilt one with auroraboot already bundled with the binary? Maybe xorriso already has some built in stuff? Maybe we should deprecate BIOS booting already (it been 20 years of EFI come on)

PR #69, opened 2024-10-22
Migrate to urfave cli v2

jimmykarily:
https://cli.urfave.org/migrate-v1-to-v2/

### DIFF
--- a/main.go
+++ b/main.go
@@ -12,12 +12,17 @@ import (
 	"github.com/urfave/cli/v2"
 )
 
+var (
+	// https://goreleaser.com/cookbooks/using-main.version/
+	version = "dev"
+)
+
 func main() {
 
 	log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stderr})
 	app := &cli.App{
 		Name:    "AuroraBoot",
-		Version: "0.1",
+		Version: version,
 		Authors: []*cli.Author{{Name: "Kairos authors", Email: "members@kairos.io"}},
 		Usage:   "auroraboot",
 		Flags: []cli.Flag{
--- a/main.go
+++ b/main.go
@@ -31,10 +31,9 @@ func main() {
 				Name: "debug",
 			},
 		},
-		Description: `
-`,
-		UsageText: ``,
-		Copyright: "kairos authors",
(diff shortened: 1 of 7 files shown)


main.go
```diff
@@ -9,33 +9,36 @@ import (
 	"github.com/kairos-io/AuroraBoot/deployer"
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
+	"github.com/urfave/cli/v2"
+)
 
-	"github.com/urfave/cli"
+var (
+	// https://goreleaser.com/cookbooks/using-main.version/
+	version = "dev"
```
Itxaka, line 17: in all other kairos artifacts we use a more especific version which also has the commit and go version. Maybe we should use the same here?
https://github.com/kairos-io/immucore/blob/main/internal/version/version.go

Itxaka (approved):

main.go
jimmykarily: That's what I went with initially but then I realized we don't use Earthly to build and release this one.I thought I should better follow the go-releaser convention. The alternative is to just pass a ldflags to goreleaser with whatever value we want. Should I change it?

jimmykarily: Actually, what you put in this `version` is up to the command that build the binary. With the changes in this PR, it's either passed as a build arg (when it's a tag) or we can pass it by hand when we build the image locally. Since we don't have an earthly target to build the binary, there is nothing else to implement, no?

jimmykarily: I pushed a commit to change this so that it uses `git describe` to calculate a version string. It still only applied to releases. The only difference is the image it builds when we push to master, which will now have a proper version like `v0.2.7-46-gd9f73da`.

jimmykarily: Checking that it works requires us to release of course. I suggest we do a patch release to test it.

Itxaka:
I dont think this is actually ok, version is version and commit is commit. We dont neccesarily want a release version to have the git version attached so thats why its stored in a different internal var on compiling

see https://github.com/kairos-io/kairos-agent/blob/main/internal/common/version.go and https://github.com/kairos-io/kairos-agent/blob/main/.goreleaser.yaml 

For master pushes, I would just push it as usual, with the master/main tag and call it a day, no need to include the git vbersion in the naming or anything, but keeping it internally stored is good in case we need to pinpoint somehow the commit it was built from in case it was wrong :D

Itxaka (approved):
still ok for me

PR #70, opened 2024-10-22
Dont require luet ot pull image

Itxaka:
This uses the sdk image pull whicyh already tries local first and doesnt require luet to be available

### DIFF
--- a/pkg/ops/iso.go
+++ b/pkg/ops/iso.go
@@ -37,7 +37,7 @@ func GenISO(name, src, dst string, i schema.ISO) func(ctx context.Context) error
 		}
 
 		log.Info().Msgf("Generating iso '%s' from '%s' to '%s'", name, src, dst)
-
+		log.Info().Msg("ENKI")
 		cfg := enkiconfig.NewBuildConfig(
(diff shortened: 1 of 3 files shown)


pkg/ops/iso.go
```diff
@@ -37,7 +37,7 @@ func GenISO(name, src, dst string, i schema.ISO) func(ctx context.Context) error
 		}
 
 		log.Info().Msgf("Generating iso '%s' from '%s' to '%s'", name, src, dst)
-
+		log.Info().Msg("ENKI")
```
jimmykarily, line 40: dev output? Otherwise that's just a technical detail, the user doesn't know (or care) what enki is.

jimmykarily (approved):

pkg/ops/iso.go
Itxaka: yes I was building aurora against the local enki and wanted to make sure I was calling the proper enki backend looool

Itxaka: done!

jimmykarily (approved):

jimmykarily (approved):

PR #77, opened 2024-10-25
Introduce build-iso command

jimmykarily:
Part of https://github.com/kairos-io/kairos/issues/1633


jimmykarily:
TODO:

- [ ] Implement "--label" flag
- [ ] Implement "--squash-no-compression" flag
- [ ] Implement "--arch" flag
- [ ] Migrate test from enki for the new command or create one if none exists

Itxaka:
> TODO:

>     * [ ]  Implement "--squash-no-compression" flag (not doing anything in enki?)

If not needed I would drop it.

Before in the old times, we were building the squash file in the livecd with compression enabled. That was very slow. So we introduced the flag to allow building with no compression.
After some time it seems like we made it the default and nobody complained. So I would ignore it and in the future drop anything related to that. we build the squash file with no compression as it only gains us a few Mbs and it speeds up building by a lot. End of case.

internal/cmd/app.go
```diff
@@ -0,0 +1,59 @@
...
+
+			if ctx.Bool("debug") {
+				internal.Log.SetLevel("debug")
+			}
+			c, r, err := ReadConfig(ctx.Args().First(), ctx.String("cloud-config"), ctx.StringSlice("set"))
+			if err != nil {
+				return err
+			}
+			c.ISO.Name = KairosDefaultArtifactName
+			c.ISO.Label = "COS_LIVE"
```
Itxaka, line 43: this should not even be a var. We need it to be called COS_LIVE IIRC, so we can just have it as a constant somewhere and just use it when needed hardcoded.

internal/cmd/build-iso.go
```diff
@@ -0,0 +1,162 @@
...
+	Usage:   "Builds an ISO from a container image or github release",
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:    "cloud-config",
+			Aliases: []string{"c"},
+			Usage:   "The cloud config to embed in the ISO",
+		},
+		&cli.StringFlag{
+			Name:    "name",
+			Aliases: []string{"n"},
```
Itxaka, line 30: put the KairosDefaultArtifactName here as a default value if nothing is passed?

internal/cmd/build-iso.go
```diff
@@ -0,0 +1,162 @@
...
+		&cli.StringFlag{
+			Name:  "overlay-uefi",
+			Usage: "Path of the overlayed uefi data",
+		},
+		&cli.StringFlag{
+			Name:  "overlay-iso",
+			Usage: "Path of the overlayed iso data",
+		},
+		&cli.StringFlag{
+			Name:  "label",
```
Itxaka, line 56: would drop this, not really over writable from the user IIRC

internal/cmd/build-iso.go
```diff
@@ -0,0 +1,162 @@
...
+			Usage: "Label of the ISO volume",
+		},
+		&cli.BoolFlag{
+			Name:  "squash-no-compression",
```
Itxaka, line 60: same, would drop this. This has to be the default

internal/cmd/build-iso.go
```diff
@@ -0,0 +1,162 @@
...
+			Value: true,
+			Usage: "Disable squashfs compression",
+		},
+		// https://github.com/kairos-io/enki/blob/6b92cbae96e92a1e36dfae2d5fdb5f3fb79bf99d/pkg/action/build-iso.go#L263
+		&cli.StringFlag{
+			Name:    "arch",
+			Aliases: []string{"a"},
```
Itxaka, line 67: woudl make the default the `runtime.GOARCH`

internal/cmd/build-iso.go
```diff
@@ -0,0 +1,162 @@
...
+			if err != nil {
+				return fmt.Errorf("reading cloud config: %w", err)
+			}
+		}
+		r := schema.ReleaseArtifact{
+			ContainerImage: source,
+		}
+		c := schema.Config{
+			ISO: schema.ISO{
+				Name:          artifactBaseName(ctx),
```
Itxaka, line 100: are we not checking beforehand that we have all the required values? This is a cli part so it should do its own validation before buildding a struct with potentially empty fields no?

pkg/ops/iso.go
```diff
@@ -31,33 +32,49 @@ func GenISO(name, src, dst string, i schema.ISO) func(ctx context.Context) error
...
+		}
+		isoLabel := enkiconstants.ISOLabel
+		if i.Label != "" {
+			isoLabel = i.Label
+		}
 		spec := &enkitypes.LiveISO{
 			RootFS:             []*v1.ImageSource{v1.NewDirSrc(src)},
 			Image:              []*v1.ImageSource{v1.NewDirSrc("/grub2"), v1.NewDirSrc(overlay)},
-			Label:              "COS_LIVE",
+			Label:              isoLabel,
```
Itxaka, line 58: as said above, this is non-negotiable :D

pkg/ops/iso.go
```diff
@@ -31,33 +32,49 @@ func GenISO(name, src, dst string, i schema.ISO) func(ctx context.Context) error
 			overlay = i.DataPath
 		}
 
+		// We are assuming StepCopyCloudConfig has already run, putting it the config in "dst"
```
Itxaka, line 35: umm this is bad. Should we check if there is a dst/config.yaml before continuing?

pkg/ops/container.go
```diff
@@ -2,15 +2,22 @@ package ops
...
 
+	"github.com/distribution/reference"
 	"github.com/kairos-io/AuroraBoot/internal"
 	sdkUtils "github.com/kairos-io/kairos-sdk/utils"
 )
 
 // PullContainerImage pulls a container image either remotely or locally from a docker daemon.
 func PullContainerImage(image, dst string) func(ctx context.Context) error {
 	return func(ctx context.Context) error {
+		_, err := reference.ParseNormalizedNamed(image)
```
Itxaka, line 16: noice

Itxaka (commented):
couple of things, mostly about things that we should drop now instead of drag around. Rest looks pretty good!

internal/cmd/app.go
jimmykarily: moved to a constant

internal/cmd/build-iso.go
jimmykarily: good point. Done.

internal/cmd/build-iso.go
jimmykarily: Yes I don't see it being overridable. Removed (also the constant I introduced for COS_LIVE in my other comment above).

internal/cmd/app.go
jimmykarily: nope, removed (see the other comment below)

internal/cmd/build-iso.go
jimmykarily: done

internal/cmd/build-iso.go
jimmykarily: done

internal/cmd/build-iso.go
jimmykarily: I do the check below (`validateISOOptions`) but you are right, it's a bit inversed. I will change to:
- create the ISO object
- validate it
- pass it to the config

jimmykarily: done

pkg/ops/iso.go
jimmykarily: yup, changed

jimmykarily: Actually, this one is needed because it's used here: https://github.com/kairos-io/enki/blob/60311c6cc4dff0ed150e79ad57a4e37b67ddf25b/pkg/action/build-iso.go#L409 and I don't see it being set to the default `enkiconstants.ISOLabel` anywhere else. I just removed the blog that allows the user to overwrite it (lines 52-54 above)

pkg/ops/iso.go
jimmykarily: I will add a check but that doesn't change the fact that we still assume the other step to have run (temporal dependency). Since steps have dependencies on other steps by definition, maybe that's ok?

jimmykarily: done

Itxaka: It may be that geniso maybe should not copy the config? Maybe we need another steps that actually copies the config only and that it's triggered by the dag dependency? Anyway, can wait

Itxaka (approved):
Good work!

PR #80, opened 2024-10-28
Drop netboot.sh script usage

Itxaka:
Use the new ExtractFileFromIso method from the sdk to extract the needed files for netbooting instead of shelling out to the netboot.sh script.

This means that we should be able to drop from the osbuilder image the netboot.sh script and the isoinfo packages


mauromorales (approved):
noiceee

PR #98, opened 2024-11-11
1633 port genkey command

jimmykarily:
Part of: https://github.com/kairos-io/kairos/issues/1633


Itxaka (approved):
Im not reading all this, this is just the genkey command dumped into here right?

jimmykarily:
> Im not reading all this, this is just the genkey command dumped into here right?

yes with some minor changes to read the args from the command `ctx` instead of `viper`.

PR #113, opened 2024-11-15
Detect version correctly so that we also push "latest" tag

jimmykarily:
Something was pushed as `latest` but not the tagged image. Not sure what:

image

### DIFF
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -32,7 +32,7 @@ jobs:
           # If the VERSION looks like a version number, assume that
           # this is the most recent version of the image and also
           # tag it 'latest'.
-          if [[ $VERSION =~ ^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}$ ]]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/image.yml
```diff
@@ -32,7 +32,7 @@ jobs:
           # If the VERSION looks like a version number, assume that
           # this is the most recent version of the image and also
           # tag it 'latest'.
-          if [[ $VERSION =~ ^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}$ ]]; then
+          if [[ $VERSION =~ ^v[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}$ ]]; then
```
jimmykarily, line 35: because `v` is part of what `git describe` returns:

```
~/workspace/kairos/AuroraBoot (also-push-latest)*$ git describe --always --tags --dirty
v0.3.2-1-g2fea056
```

Itxaka:
I think master is whats pushed to latest by default

Itxaka (approved):

PR #122, opened 2024-11-25
Do not change name to extend-cmdline output files

Itxaka:
For some weird reason we were changing the name for the extended cmdline efi output file, which is worn. It wont collide with anything as we are just generating a single file, so there is no need to change its name.

This generate weird names like norole_EXTENDEDCMDLINE.efi and conf files which then did not pass the upgrade requirements (which checks for norole.efi)

### DIFF
--- a/internal/cmd/build-uki.go
+++ b/internal/cmd/build-uki.go
@@ -936,18 +936,17 @@ func createContainer(sourceDir, outputDir, artifactName, version string, logger 
 // GetUkiCmdline returns the cmdline to be used for the kernel.
 // The cmdline can be overridden by the user using the cmdline flag.
 // For each cmdline passed, we generate a uki file with that cmdline
-// extend-cmdline will just extend the default cmdline so we only create one efi file
-// extra-cmdline will create a new efi file for each cmdline passed
+// extend-cmdline will just extend the default cmdline so we only create one efi file. Artifact name is the default one
+// extra-cmdline will create a new efi file for each cmdline passed. artifact name is generated from the cmdline
 func GetUkiCmdline(cmdlineExtend, bootBranding string, extraCmdlines []string) []utils.BootEntry {
 	defaultCmdLine := constants.UkiCmdline + " " + constants.UkiCmdlineInstall
 
-	// Extend only
+	// Override the default cmdline if the user passed one
 	if cmdlineExtend != "" {
-		cmdline := defaultCmdLine + " " + cmdlineExtend
 		return []utils.BootEntry{{
-			Cmdline:  cmdline,
+			Cmdline:  defaultCmdLine + " " + cmdlineExtend,
 			Title:    bootBranding,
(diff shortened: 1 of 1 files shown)


internal/cmd/build-uki.go
```diff
@@ -936,26 +936,21 @@ func createContainer(sourceDir, outputDir, artifactName, version string, logger
...
-	// Extend only
+	// Override the default cmdline if the user passed one
 	if cmdlineExtend != "" {
-		cmdline := defaultCmdLine + " " + cmdlineExtend
-		return []utils.BootEntry{{
-			Cmdline:  cmdline,
-			Title:    bootBranding,
-			FileName: NameFromCmdline(constants.ArtifactBaseName, cmdline),
-		}}
+		defaultCmdLine = defaultCmdLine + " " + cmdlineExtend
```
jimmykarily, line 946: with this change, the extraCmdline on line 957, will also get the extended one, which wasn't the case up to now. I'm not sure if that's what we want or not. I'm just saying it's changing the behavior.

internal/cmd/build-uki.go
```diff
@@ -936,26 +936,21 @@ func createContainer(sourceDir, outputDir, artifactName, version string, logger
...
 	}
 
 	// default entry
 	result := []utils.BootEntry{{
 		Cmdline:  defaultCmdLine,
 		Title:    bootBranding,
-		FileName: NameFromCmdline(constants.ArtifactBaseName, defaultCmdLine),
+		FileName: constants.ArtifactBaseName,
```
jimmykarily, line 953: if this is named "norole.efi" then the extraCmdlines below would be called "norole_thingshere.efi" and such, no? Will these names work when we install? What is `norole` being replaced with in that case?

Itxaka: in any case, this change in this line is moot because the result is the same, in both cases due to the default cmdline, the artifact name will be norole always. I just wanted this to be more clear, but can revert it if needed.

internal/cmd/build-uki.go
```diff
@@ -936,26 +936,25 @@ func createContainer(sourceDir, outputDir, artifactName, version string, logger
...
-	// Extend only
+	// Override the default cmdline if the user passed one
 	if cmdlineExtend != "" {
-		cmdline := defaultCmdLine + " " + cmdlineExtend
 		return []utils.BootEntry{{
-			Cmdline:  cmdline,
+			Cmdline:  defaultCmdLine + " " + cmdlineExtend,
 			Title:    bootBranding,
-			FileName: NameFromCmdline(constants.ArtifactBaseName, cmdline),
+			FileName: constants.ArtifactBaseName,
```
Itxaka, line 949: this is the real change, for extended cmdlines, we want to have the artifacts named norole ALWAYS. It makes no sense to change the name because its gonna be the only artifacts in there.

jimmykarily (approved):

PR #125, opened 2024-11-27
Switch to Fedora for the auroraboot image

jimmykarily:
because opensuse/leap doesn't have latest systemd and older systemd-repart doesn't support the flags we are passing in the sysext command (e.g.  --make-ddi).

Our other option would be to switch to Tumbleweed but that might be too unstable for us to use for Auroraboot.

This is reviving the work here:
https://github.com/kairos-io/osbuilder/pull/128

later reverted here:
https://github.com/kairos-io/osbuilder/commit/7813c2eec45a1db8ff9b1283ed2c9fa73e4cf878

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
+ARG FEDORA_VERSION=40
 ARG LUET_VERSION=0.35.5
-ARG LEAP_VERSION=15.5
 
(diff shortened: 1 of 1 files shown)


jimmykarily:
We need to test building an arm image with this

Itxaka (approved):

PR #129, opened 2024-12-04
Create directories with 755 permissions

jimmykarily:
because otherwise the rootfs `/` has permissions 755 which results in all users except "root" to not be able to login or run any commands

### DIFF
--- a/deployer/steps.go
+++ b/deployer/steps.go
@@ -12,7 +12,7 @@ import (
 func (d *Deployer) StepPrepNetbootDir() error {
 	return d.Add(opPrepareNetboot, herd.WithCallback(
 		func(ctx context.Context) error {
-			return os.MkdirAll(d.dstNetboot(), 0700)
+			return os.MkdirAll(d.dstNetboot(), 0755)
 		},
 	))
 }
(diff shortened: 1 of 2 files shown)


pkg/utils/fs.go
```diff
@@ -72,7 +72,7 @@ func TempDir(fs v1.FS, dir, prefix string) (name string, err error) {
 	}
 	// This skips adding random stuff to the created temp dir so the temp dir created is predictable for testing
 	if _, isTestFs := fs.(*vfst.TestFS); isTestFs {
-		err = MkdirAll(fs, filepath.Join(dir, prefix), 0700)
+		err = MkdirAll(fs, filepath.Join(dir, prefix), 0755)
```
Itxaka, line 75: I think this one is not necessary as it's only for testing IIRC

jimmykarily: but doesn't hurt either, right? Better be consistent.

deployer/steps.go
```diff
@@ -12,22 +12,22 @@ import (
 func (d *Deployer) StepPrepNetbootDir() error {
 	return d.Add(opPrepareNetboot, herd.WithCallback(
 		func(ctx context.Context) error {
-			return os.MkdirAll(d.dstNetboot(), 0700)
+			return os.MkdirAll(d.dstNetboot(), 0755)
```
Itxaka, line 15: we should probably use a constant for this as we do on other places, but lets fix it first

PR #131, opened 2024-12-05
Allow spaces before '#'

jimmykarily:
to make this work:

https://github.com/kairos-io/AuroraBoot/blob/243b444685b2cc152506c07cdb6a07c2b6e5bdb6/luet-amd64.yaml#L15


mauromorales (approved):

Itxaka (approved):
wow, good catch!

PR #132, opened 2024-12-05
Bring back the earthly target to update luet repos

jimmykarily:
because renovate doesn't seem to understand our image versioning

It was first introduced here:

https://github.com/kairos-io/AuroraBoot/commit/cc8b857ec7c483c63182659c96c6f7caa0fccd0c#diff-2705f6016a64189abd14041e2f3e8dbb829553295d9088e4eb42eccb5e61bf47R84

(when we migrated it from enki)

Also update the luet repos using the new target


Itxaka (changes requested):
I don't get why, renovate should update them as it does on the framework?

Itxaka (approved):

PR #134, opened 2024-12-09
feat(no UKI): improve EFI booting from USB stick

mudler:
Fixes: https://github.com/kairos-io/kairos/issues/3063

https://wiki.debian.org/RepackBootableISO

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -112,6 +112,7 @@ func GetXorrisoBooloaderArgs(root string) []string {
 		"-boot_image", "any", "efi_path=--interval:appended_partition_2:all::",
 		"-boot_image", "any", "platform_id=0xef",
 		"-boot_image", "any", "emul_type=no_emulation",
+		"-boot_image", "grub", "partition_entry=gpt_basdat",
 	}
 	return args
 }


pkg/constants/constants.go
```diff
@@ -112,6 +112,7 @@ func GetXorrisoBooloaderArgs(root string) []string {
 		"-boot_image", "any", "efi_path=--interval:appended_partition_2:all::",
 		"-boot_image", "any", "platform_id=0xef",
 		"-boot_image", "any", "emul_type=no_emulation",
+		"-isohybrid-gpt-basdat",
```
jimmykarily, line 115: https://wiki.syslinux.org/wiki/index.php?title=Isohybrid

I'm not sure I understand what I'm reading :D. How can we test if this works or not?

jimmykarily (approved):

pkg/constants/constants.go
Itxaka: I guess uki tests would cover this as htey are EFI only, but we never had issues. But its troue that people at SpectroCloud tested with normal (non-uki) isos and had problems booting them from some EFI firmwares and adding this helped them boot the isos, and we got no feeback of breaking anything on BIOS boots so....I guess its tested already, just making the hybrid isos more compatible.

What we should do is NOT generate hybrid isos for UKI stuff as there is no need for that, but that requires heavier changes to the process.

Itxaka (approved):

pkg/constants/constants.go
Itxaka:
```
Mark the current El Torito boot image (see options -b and -e) in an actually invalid GPT as partition of type Basic Data. This works only with -isohybrid-mbr and has the same impact on the system area as -efi-boot-part. It cannot be combined with -efi-boot-part or -hfsplus.
The first three boot images which are marked by GPT will also show up as partition entries in MBR. The MBR partition of type 0xEF is what actually is used by EFI firmware for booting from USB stick. The MBR partition for PC-BIOS gets type 0x00 rather than 0x17 in this case. Often the further MBR entries are the ones which actually get used by EFI.
```

So the proper option under xorriso is `"-boot_image", "any", "efi_boot_part=-efi-boot-image"`

yes, the flag value is `-efi-boot-image` dont ask me why.

Itxaka:
```
-efi-boot-part disk_path

Copy a file from disk into the emerging ISO image and mark it by a GPT entry as EFI System Partition. EFI boot firmware is supposed to use a FAT filesystem image in such a partition for booting from USB stick or hard disk.
Instead of a disk_path, the word --efi-boot-image may be given. It exposes in GPT the content of the first El Torito EFI boot image as EFI system partition. EFI boot images are introduced by options -e or --efi-boot. The affected EFI boot image cannot show up in HFS+ because it is stored outside the HFS+ partition.
```

mudler: mmm isn't "-boot_image isolinux partition_entry=gpt_basdat"  as well needed?  https://wiki.debian.org/RepackBootableISO

mudler:
Closing in favor of https://github.com/kairos-io/AuroraBoot/pull/135

PR #135, opened 2024-12-09
Fix different behaviour between usb and cd in efi

Itxaka:
Seems that the actual issue was not that the USB is not recognized, but the usb boots and it gets a grub shell because it cannot load the default grub.cfg for efi.

This seems to be a side issue of how efi booting works differently from livecd or usb in grub.
Looks like when booting from livecd, the efi artifacts are loaded from the efi image file BUT the livecd is mounted as root for grub.
When booting from usb it seems like the efi artifacts are loaded from the efi image AND the efi image is mounted as root for grub.

In both cases, grub is hardcoded to search for the grub config in /EFI/boot/grub.cfg (so alongside the grubx64.efi file)
When we build the ISO we only copy those artifacts on the ISO root, so you got the /EFI/boot/grub.cfg directly accessible when booting from livecd.
But when booting from usb the livecd is ignored and only the contents of the uefi.img are in the root, so /EFI/boot/grub.cfg does not exists.

This is fixed by copying it to both places so it works on both boot methods.

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -4,7 +4,6 @@ import (
 	_ "embed"
 	"fmt"
 	"os"
-	"path/filepath"
 )
 
 // Eltorito image is basically a grub cdboot.img + grub core image
@@ -107,11 +106,11 @@ func GetXorrisoBooloaderArgs(root string) []string {
 		"-boot_image", "any", "platform_id=0x00",
 		"-boot_image", "any", "emul_type=no_emulation",
 		"-boot_image", "any", "load_size=2048",
-		"-append_partition", "2", "0xef", filepath.Join(root, IsoEFIPath),
 		"-boot_image", "any", "next",
-		"-boot_image", "any", "efi_path=--interval:appended_partition_2:all::",
 		"-boot_image", "any", "platform_id=0xef",
 		"-boot_image", "any", "emul_type=no_emulation",
+		"-boot_image", "any", "efi_path=" + IsoEFIPath, // Tell where the efi path is
+		"-boot_image", "any", "efi_boot_part=--efi-boot-image", // Tell where the efi path is for booting
 	}
 	return args
 }
--- a/pkg/ops/iso.go
+++ b/pkg/ops/iso.go
@@ -397,7 +397,7 @@ func (b BuildISOAction) createEFI(rootdir string, isoDir string) error {
 	// the uefi.img is loaded into memory and run, but grub only sees the livecd root
 	// WARNING: If we load the image into an usb stick, grub will not use the livecd root as the grub root, so it cannot find the grub.cfg
 	// So we copy in 2 places, into the livecd and into the img
-	
+
 	// This is used when booting from usb
 	if err = b.writeDefaultGrubEfiCfg(temp); err != nil {
 		b.cfg.Logger.Errorf("Failed writing grub.cfg: %v", err)
@@ -424,16 +424,6 @@ func (b BuildISOAction) createEFI(rootdir string, isoDir string) error {
 	b.cfg.Logger.Infof("Detected Flavor: %s", flavor)
 	if strings.Contains(strings.ToLower(flavor), "ubuntu") {
 		b.cfg.Logger.Infof("Ubuntu based ISO detected, copying grub.cfg to /EFI/ubuntu/grub.cfg")
-		err = utils.MkdirAll(b.cfg.Fs, filepath.Join(temp, "EFI/ubuntu/"), constants.DirPerm)
-		if err != nil {
-			b.cfg.Logger.Errorf("Failed writing grub.cfg: %v", err)
-			return err
-		}
-		err = utils.MkdirAll(b.cfg.Fs, filepath.Join(isoDir, "EFI/ubuntu/"), constants.DirPerm)
-		if err != nil {
-			b.cfg.Logger.Errorf("Failed writing grub.cfg: %v", err)
-			return err
-		}
 
 		if err = b.writeUbuntuGrubEfiCfg(temp); err != nil {
 			b.cfg.Logger.Errorf("Failed writing grub.cfg: %v", err)
@@ -485,15 +475,19 @@ func (b BuildISOAction) createEFI(rootdir string, isoDir string) error {
 
 // writeDefaultGrubEfiCfg writes the default grub.cfg for the EFI image in teh given path
 func (b *BuildISOAction) writeDefaultGrubEfiCfg(path string) error {
-        calculatedPath := filepath.Join(path, constants.EfiBootPath, constants.GrubCfg)
+	calculatedPath := filepath.Join(path, constants.EfiBootPath, constants.GrubCfg)
 	return b.cfg.Fs.WriteFile(calculatedPath, []byte(constants.GrubEfiCfg), constants.FilePerm)
 }
 
 func (b *BuildISOAction) writeUbuntuGrubEfiCfg(path string) error {
-        calculatedPath := filepath.Join(path, "EFI/ubuntu/", constants.GrubCfg)
+	calculatedPath := filepath.Join(path, "EFI/ubuntu/", constants.GrubCfg)
+	err := utils.MkdirAll(b.cfg.Fs, calculatedPath, constants.DirPerm)
+	if err != nil {
+		b.cfg.Logger.Errorf("Failed writing grub.cfg: %v", err)
+		return err
+	}
 	return b.cfg.Fs.WriteFile(calculatedPath, []byte(constants.GrubEfiCfg), constants.FilePerm)
 }
-
 
 // copyShim copies the shim files into the EFI partition
 // tempdir is the temp dir where the EFI image is generated from
--- a/pkg/ops/iso.go
+++ b/pkg/ops/iso.go
@@ -397,13 +397,14 @@ func (b BuildISOAction) createEFI(rootdir string, isoDir string) error {
 	// the uefi.img is loaded into memory and run, but grub only sees the livecd root
 	// WARNING: If we load the image into an usb stick, grub will not use the livecd root as the grub root, so it cannot find the grub.cfg
 	// So we copy in 2 places, into the livecd and into the img
-	err = b.cfg.Fs.WriteFile(filepath.Join(temp, constants.EfiBootPath, constants.GrubCfg), []byte(constants.GrubEfiCfg), constants.FilePerm)
-	if err != nil {
+	
+	// This is used when booting from usb
+	if err = b.writeDefaultGrubEfiCfg(filepath.Join(temp, constants.EfiBootPath, constants.GrubCfg)); err != nil {
 		b.cfg.Logger.Errorf("Failed writing grub.cfg: %v", err)
 		return err
 	}
-	err = b.cfg.Fs.WriteFile(filepath.Join(isoDir, constants.EfiBootPath, constants.GrubCfg), []byte(constants.GrubEfiCfg), constants.FilePerm)
-	if err != nil {
+	// This is used when booting from cdrom
+	if err = b.writeDefaultGrubEfiCfg(filepath.Join(isoDir, constants.EfiBootPath, constants.GrubCfg)); err != nil {
 		b.cfg.Logger.Errorf("Failed writing grub.cfg: %v", err)
 		return err
 	}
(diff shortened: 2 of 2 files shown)


pkg/constants/constants.go
```diff
@@ -107,11 +106,11 @@ func GetXorrisoBooloaderArgs(root string) []string {
 		"-boot_image", "any", "platform_id=0x00",
 		"-boot_image", "any", "emul_type=no_emulation",
 		"-boot_image", "any", "load_size=2048",
-		"-append_partition", "2", "0xef", filepath.Join(root, IsoEFIPath),
 		"-boot_image", "any", "next",
-		"-boot_image", "any", "efi_path=--interval:appended_partition_2:all::",
 		"-boot_image", "any", "platform_id=0xef",
 		"-boot_image", "any", "emul_type=no_emulation",
+		"-boot_image", "any", "efi_path=" + IsoEFIPath, // Tell where the efi path is
```
Itxaka, line 112: this is just an improvement over what we had.
Before:
 - uefi.img in the disc
 - second partition in the disc with uefi contents

Now:
 - just the uefi.img in the disc

pkg/ops/iso.go
```diff
@@ -394,6 +395,13 @@ func (b BuildISOAction) createEFI(rootdir string, isoDir string) error {
 	// Its read from the root of the livecd, so we need to copy it into /EFI/BOOT/grub.cfg
 	// This is due to the hybrid bios/efi boot mode of the livecd
 	// the uefi.img is loaded into memory and run, but grub only sees the livecd root
+	// WARNING: If we load the image into an usb stick, grub will not use the livecd root as the grub root, so it cannot find the grub.cfg
+	// So we copy in 2 places, into the livecd and into the img
+	err = b.cfg.Fs.WriteFile(filepath.Join(temp, constants.EfiBootPath, constants.GrubCfg), []byte(constants.GrubEfiCfg), constants.FilePerm)
+	if err != nil {
+		b.cfg.Logger.Errorf("Failed writing grub.cfg: %v", err)
+		return err
+	}
```
mauromorales, line 404: could we extract this into a function and just pass it once? I never know what's the best in Go, I find it so verbose, but I know it's common to do this, but it's quite hard for me to see from these lines and the ones below that the only difference is temp and isoDir

pkg/ops/iso.go
```diff
@@ -415,11 +423,21 @@ func (b BuildISOAction) createEFI(rootdir string, isoDir string) error {
...
+		if err != nil {
+			b.cfg.Logger.Errorf("Failed writing grub.cfg: %v", err)
+			return err
+		}
 		err = utils.MkdirAll(b.cfg.Fs, filepath.Join(isoDir, "EFI/ubuntu/"), constants.DirPerm)
 		if err != nil {
 			b.cfg.Logger.Errorf("Failed writing grub.cfg: %v", err)
 			return err
 		}
+		err = b.cfg.Fs.WriteFile(filepath.Join(temp, "EFI/ubuntu/", constants.GrubCfg), []byte(constants.GrubEfiCfg), constants.FilePerm)
```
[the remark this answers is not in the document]
Itxaka: This is a nice trivia.

So when you build grub efi, you can set where it looks for the config rigth?

Most "normal" distros, are not stupid, so they set the grub.cfg search in the same dir as the grub efi file, so potentially you could have different grub versions with different grub configs.

/EFI/grub2/grub2.efi which will read the cfg file at /EFI/grub2/grub.cfg
/EFI/grub2-beta/grub2.efi which will read the cfg file at /EFI/grub2-beta/grub.cfg


BUT! Ubuntu is special, because why not?
So they hardcode where the grub.cfg file is and set a generic path in which to search for it, which is /EFI/ubuntu/grub.cfg

So we have to account for that and make sure that if the distro is ubuntu, we have to set the grub at that specific location.......yeah.

pkg/ops/iso.go
Itxaka: I think the comment above should be good enough no? even if we set it to a function, the call would be as cryptic as is currently now no?

```
err := WriteGrubCfg(tmpdir)
err = WriteGrubCfg(isodir)
```

pkg/ops/iso.go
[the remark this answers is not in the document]
mauromorales: 🤦 ubuntu

thanks for clarifying

pkg/ops/iso.go
mauromorales: for me something like

```
err := WriteGrubCfg(temp)
err = WriteGrubCfg(isodir)
```

is always clearer because it really shows only the different arguments passed and not the calculation of paths like here:

```
err = b.cfg.Fs.WriteFile(filepath.Join(temp, constants.EfiBootPath, constants.GrubCfg), []byte(constants.GrubEfiCfg), constants.FilePerm)	
err = b.cfg.Fs.WriteFile(filepath.Join(isoDir, constants.EfiBootPath, constants.GrubCfg), []byte(constants.GrubEfiCfg), constants.FilePerm)	
```

It's simply a lower mental load when reading the code, at least for me coming from the OO world, because you avoid having to also perform these calculations 

```
filepath.Join(temp, constants.EfiBootPath, constants.GrubCfg), []byte(constants.GrubEfiCfg), constants.FilePerm
```

as part of the arguments and if you are interested in them, then you jump into the function making them

Itxaka: oki, will do, one sec

Itxaka: how about now?

Itxaka: oh I think I messed up here, you meant filling the path in the function? But thats just moving the complexity into a different part of the code.

Here in fact, it makes the code more complex because then you have to have 2 different functions for writing the generic config file one for generic, one for ubuntu as their paths are different.
And the functions are just a wrapper for a call to writefile.

I dunno, for me this is worst, having extra functions for a one line makes having to jump around to the function to see what it does. But I dont care either way, I can do it if you want :D

pkg/ops/iso.go
```diff
@@ -397,13 +397,14 @@ func (b BuildISOAction) createEFI(rootdir string, isoDir string) error {
 	// the uefi.img is loaded into memory and run, but grub only sees the livecd root
 	// WARNING: If we load the image into an usb stick, grub will not use the livecd root as the grub root, so it cannot find the grub.cfg
 	// So we copy in 2 places, into the livecd and into the img
-	err = b.cfg.Fs.WriteFile(filepath.Join(temp, constants.EfiBootPath, constants.GrubCfg), []byte(constants.GrubEfiCfg), constants.FilePerm)
-	if err != nil {
+	
+	// This is used when booting from usb
+	if err = b.writeDefaultGrubEfiCfg(filepath.Join(temp, constants.EfiBootPath, constants.GrubCfg)); err != nil {
```
mauromorales, line 402:
```suggestion
	if err = b.writeDefaultGrubEfiCfg(temp); err != nil {
```

pkg/ops/iso.go
```diff
@@ -397,13 +397,14 @@ func (b BuildISOAction) createEFI(rootdir string, isoDir string) error {
...
 		b.cfg.Logger.Errorf("Failed writing grub.cfg: %v", err)
 		return err
 	}
-	err = b.cfg.Fs.WriteFile(filepath.Join(isoDir, constants.EfiBootPath, constants.GrubCfg), []byte(constants.GrubEfiCfg), constants.FilePerm)
-	if err != nil {
+	// This is used when booting from cdrom
+	if err = b.writeDefaultGrubEfiCfg(filepath.Join(isoDir, constants.EfiBootPath, constants.GrubCfg)); err != nil {
```
mauromorales, line 407:
```suggestion
	if err = b.writeDefaultGrubEfiCfg(isoDir); err != nil {
```

pkg/ops/iso.go
```diff
@@ -433,13 +434,12 @@ func (b BuildISOAction) createEFI(rootdir string, isoDir string) error {
 			b.cfg.Logger.Errorf("Failed writing grub.cfg: %v", err)
 			return err
 		}
-		err = b.cfg.Fs.WriteFile(filepath.Join(temp, "EFI/ubuntu/", constants.GrubCfg), []byte(constants.GrubEfiCfg), constants.FilePerm)
-		if err != nil {
+
+		if err = b.writeDefaultGrubEfiCfg(filepath.Join(temp, "EFI/ubuntu/", constants.GrubCfg)); err != nil {
```
mauromorales, line 438:
```suggestion
		if err = b.writeUbuntuGrubEfiCfg(temp); err != nil {
```

pkg/ops/iso.go
```diff
@@ -433,13 +434,12 @@ func (b BuildISOAction) createEFI(rootdir string, isoDir string) error {
...
 			b.cfg.Logger.Errorf("Failed writing grub.cfg: %v", err)
 			return err
 		}
-		err = b.cfg.Fs.WriteFile(filepath.Join(isoDir, "EFI/ubuntu/", constants.GrubCfg), []byte(constants.GrubEfiCfg), constants.FilePerm)
-		if err != nil {
+		if err = b.writeDefaultGrubEfiCfg(filepath.Join(isoDir, "EFI/ubuntu/", constants.GrubCfg)); err != nil {
```
mauromorales, line 442:
```suggestion
		if err = b.writeUbuntuGrubEfiCfg(isoDir); err != nil {
```

pkg/ops/iso.go
```diff
@@ -483,6 +483,11 @@ func (b BuildISOAction) createEFI(rootdir string, isoDir string) error {
 	return nil
 }
 
+// writeDefaultGrubEfiCfg writes the default grub.cfg for the EFI image in teh given path
+func (b *BuildISOAction) writeDefaultGrubEfiCfg(path string) error {
+	return b.cfg.Fs.WriteFile(path, []byte(constants.GrubEfiCfg), constants.FilePerm)
+}
```
mauromorales, line 489:
```suggestion
func (b *BuildISOAction) writeDefaultGrubEfiCfg(path string) error {
        calculatedPath := filepath.Join(path, constants.EfiBootPath, constants.GrubCfg)
	return b.cfg.Fs.WriteFile(calculatedPath, []byte(constants.GrubEfiCfg), constants.FilePerm)
}

func (b *BuildISOAction) writeUbuntuGrubEfiCfg(path string) error {
        calculatedPath := filepath.Join(path, "EFI/ubuntu/", constants.GrubCfg)
	return b.cfg.Fs.WriteFile(calculatedPath, []byte(constants.GrubEfiCfg), constants.FilePerm)
}

```

mauromorales (commented):
I mean more like with the current comments I added, because it encapsulates the logic of the "hard coded" ubuntu logic and also allows commenting why it is needed in a single place.

pkg/ops/iso.go
```diff
@@ -415,16 +424,25 @@ func (b BuildISOAction) createEFI(rootdir string, isoDir string) error {
 	b.cfg.Logger.Infof("Detected Flavor: %s", flavor)
 	if strings.Contains(strings.ToLower(flavor), "ubuntu") {
 		b.cfg.Logger.Infof("Ubuntu based ISO detected, copying grub.cfg to /EFI/ubuntu/grub.cfg")
-		err = utils.MkdirAll(b.cfg.Fs, filepath.Join(isoDir, "EFI/ubuntu/"), constants.DirPerm)
+		err = utils.MkdirAll(b.cfg.Fs, filepath.Join(temp, "EFI/ubuntu/"), constants.DirPerm)
```
mauromorales, line 427: is this a mistake of mine? feels like it should still be isoDir, not sure

Also I guess this can be moved into the function

Itxaka: no, its correct, we need to create that dir for sure in both places anyway. Isodir creation is just below

pkg/ops/iso.go
```diff
@@ -465,6 +483,18 @@ func (b BuildISOAction) createEFI(rootdir string, isoDir string) error {
...
 }
 
+// writeDefaultGrubEfiCfg writes the default grub.cfg for the EFI image in teh given path
+func (b *BuildISOAction) writeDefaultGrubEfiCfg(path string) error {
+        calculatedPath := filepath.Join(path, constants.EfiBootPath, constants.GrubCfg)
+	return b.cfg.Fs.WriteFile(calculatedPath, []byte(constants.GrubEfiCfg), constants.FilePerm)
+}
+
+func (b *BuildISOAction) writeUbuntuGrubEfiCfg(path string) error {
+        calculatedPath := filepath.Join(path, "EFI/ubuntu/", constants.GrubCfg)
```
Itxaka, line 493:
```suggestion
        calculatedPath := filepath.Join(path, "EFI/ubuntu/", constants.GrubCfg)
        err = utils.MkdirAll(b.cfg.Fs, calculatedPath), constants.DirPerm)
        if err != nil { return err }
        
```

pkg/ops/iso.go
Itxaka: I moved it into the function

mauromorales (approved):

PR #139, opened 2024-12-17
Implement raw disk image generation in go

Itxaka:
Drops the use of the script to generate it on pure golang Also drops some packages which are no longer needed from luet as we use the artifacts from the rootfs to support secureboot


pkg/constants/constants.go
```diff
@@ -54,6 +54,11 @@ const (
 		"\nset prefix=($root)" + GrubPrefixDir +
 		"\nconfigfile $prefix/" + GrubCfg
 
+	// GrubEfiRecovery Used for RAW images as we chainload the grub config in the recovery partition
+	GrubEfiRecovery = "search --no-floppy --label --set=root COS_RECOVERY" +
+		"\nset root=($root)" +
+		"\nset prefix=($root)/grub2\n" +
+		"configfile ($root)/etc/cos/grub.cfg"
```
mauromorales, line 61:
```suggestion
		"\nset prefix=($root)/grub2" +
		"\nconfigfile ($root)/etc/cos/grub.cfg"
```

pkg/ops/rawDiskGeneration.go
```diff
@@ -0,0 +1,700 @@
...
+	if err != nil {
+		internal.Log.Logger.Error().Err(err).Str("target", filepath.Join(tmpDirOem, resetCloudInit)).Msg("failed to write cloud config")
+		return "", err
+	}
+
+	OemPartitionImage := v1.Image{
+		File:       filepath.Join(r.tmpDir, "oem.img"),
+		FS:         agentConstants.LinuxImgFs,
+		Label:      agentConstants.OEMLabel,
+		Size:       64,
```
jimmykarily, line 175: move to a constant?

pkg/ops/rawDiskGeneration.go
```diff
@@ -0,0 +1,700 @@
...
+	recoveryImage.Size = uint(size)
+
+	_, err = r.elemental.DeployImage(&recoveryImage, false)
+	// Create recovery.squash from the rootfs into the recovery partition under cOS/
+	if err != nil {
+		internal.Log.Logger.Error().Err(err).Str("source", r.Source).Interface("image", recoveryImage).Msg("failed to create recovery image")
+		return "", err
+	}
+
+	// TODO: Copy the grub artifacts to the recovery partition under grub2/ so they can be used by the EFI/BIOS grub
```
jimmykarily, line 229: is this what the code below does?

pkg/ops/rawDiskGeneration.go
```diff
@@ -0,0 +1,700 @@
...
+			return err
+		}
+	} else {
+		// Before 3.2.x the kairos info was in /etc/os-release
+		label, err = sdkUtils.OSRelease("IMAGE_LABEL", filepath.Join(r.Source, "etc/os-release"))
+		if err != nil {
+			internal.Log.Logger.Error().Err(err).Msg("failed to get image label")
+			return err
+		}
+	}
```
jimmykarily, line 405: this can be extracted to a function

pkg/ops/rawDiskGeneration.go
```diff
@@ -0,0 +1,700 @@
...
+		internal.Log.Logger.Error().Err(err).Str("target", rawDiskFile).Msg("failed to truncate final disk")
+		return err
+	}
+
+	// Create a GPT partition table
+	// Size needs to be rounded to the nearest 512 bytes as that's the sector size
+	// Leave 1MB at the start for alignment+partition table (so start at sector 2048)
+	// 2MB for the BIOS boot partition in case we want to do hybrid boot
+	// Which would mean installing grub to the BIOS boot partition
+	// BIOS BOOT Legacy, currently does nothing
```
jimmykarily, line 552: omg! Good job figuring all this out!

jimmykarily (commented):
Some comments may not be actionable on this PR. Feel free to comment and skip.
Also, my brain got tired near the end, I might have missed things. But we should have a test for this soon, so...

Good job!

pkg/ops/rawDiskGeneration.go
```diff
@@ -0,0 +1,700 @@
...
+// What it does
+// Create images for all three partitions
+// EFI, OEM, Recovery
+// EFI is a fat32 partition with the grub artifacts
+// OEM is an ext2 partition with the cloud config
+// Recovery is an ext2 partition with the rootfs in an image file called recovery.img under cOS/
+// Mount them with loop devices and copy the artifacts to them
+// Once the images are created, they are concatenated together with a GPT partition table
+// The final image is then truncated to the nearest sector size
+// TODO: Add BIOS support
```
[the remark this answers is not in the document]
Itxaka: nah, thats just a shim that does nothing :D

pkg/ops/rawDiskGeneration.go
```diff
@@ -0,0 +1,700 @@
...
+
+type RawImage struct {
+	CloudConfig string               // cloud config to copy to the oem partition, if none provided a default one will be created with the kairos user
+	Source      string               // Source image to copy the artifacts from, which will be the rootfs in the final image
+	Output      string               // Output image destination dir. Final image name will be based on the contents of the source /etc/kairos-release file
+	FinalSize   uint64               // Final size of the disk image in MB
+	tmpDir      string               // A temp dir to do all work on
+	elemental   *elemental.Elemental // Elemental instance to use for the operations
+	efi         bool                 // If the image should be EFI or BIOS
+	config      *config.Config       // config to use for the operations
```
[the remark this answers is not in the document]
Itxaka: yesh, it was easy to just get a config method with the defaults `config.NewConfig(config.WithLogger(internal.Log))` so when we need to pass it along, is filled.

> As far as I can tell, we use only Fs and Logger from this object in 99% of the cases. There is only a a couple of calls to GetSourceSize which [is using ImageExtractor and Platform](https://github.com/kairos-io/kairos-agent/blob/1197f49c006b23f56c29782ae7a2fcbfde7759bb/pkg/config/spec.go#L862) from it.

This does more that you think of, for example, passes the proper log down the line to the runner, the yip runner and so on, so everything is contained in there. It calls the kernel mount, it calls mkfs with the runner, etc.. so its doing a LOT more that what it looks like :D

> [Config](https://github.com/kairos-io/kairos-agent/blob/1197f49c006b23f56c29782ae7a2fcbfde7759bb/pkg/config/config.go#L129-L160) is a store-everything beast that every time we use it, it opens a world of opportunities to the functions to do whatever they want (without even caring if the Config object is fully populated or not).

I would not say that its true. Config is a struct that contains a set of know interfaces that we use along the projects, like the Runner or the FS. This is a good way of standarizing access to those resources everywhere and knowing we can call the same stuff everywhere AND mock things if needed. i.e. all of our functions can use Config.Fs.Writefile and dont care if its a real FS or a test one in memory.

Not only that, but we can take advantage of the same functions that all take a FS for example.

If the Config object is indeed empty, then its the caller responsibility to fill it up, we provide a nice entrypoint to get a Fullfilled Config object easily.


Just using Config here, means that I can add testing to everything here without worrying about if Im gonna break the system where the test is running or Im gonna really call teh kernel to mount stuff.

Yes, its a bit more complex, but I think the gains on stadarization across projects and the testing capabilities totally won in this case.

pkg/ops/rawDiskGeneration.go
```diff
@@ -0,0 +1,700 @@
...
+	err := fsutils.MkdirAll(r.config.Fs, tmpDirOem, 0755)
+	defer r.config.Fs.RemoveAll(tmpDirOem)
+
+	// This is where the oem partition will be mounted to copy the files to
+	tmpDirOemMount := filepath.Join(r.tmpDir, "oem-mount")
+	err = fsutils.MkdirAll(r.config.Fs, tmpDirOemMount, 0755)
+	defer r.config.Fs.RemoveAll(tmpDirOemMount)
+
+	// Copy the cloud config to the oem partition if htere is any
+	if r.CloudConfig != "" {
```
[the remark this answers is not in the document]
Itxaka: yup, I forgot I think, this should come from the Aurora call directly. A bit confusing how or where, I think Aurora puts the config somewhere already due to a previous step in the DAG? Gonna check it

pkg/ops/rawDiskGeneration.go
Itxaka: Im trying to understand where that fixed value came from in the script :D Maybe its a constant already and didnt notice, but yeah, will do

pkg/ops/rawDiskGeneration.go
Itxaka: no, this is still missing the grub2 BIOS artifacts I think. EFI does not matter here ans it has the bundled artifacts in the efi file but BIOS grub will look up the mods and such (the usual /boot/grub/x86-64/loopback.mod for example)

So this needs a look up when implementing the bios part :(

pkg/ops/rawDiskGeneration.go
```diff
@@ -0,0 +1,700 @@
...
+
+	return efiPartitionImage.File, nil
+}
+
+func (r *RawImage) Build() error {
+	var bootImagePath string
+	var err error
+
+	// If we dont have a tempdir, create one
+	if r.tmpDir == "" {
```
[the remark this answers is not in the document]
Itxaka: I think I had this due to the methods being public VS private. 

So I though the Build as the entrypoint for the RawImage creation, which made sense. So we setup the tmpdir here and then its used everywhere.

But then I though that for the nvidia images it may be necessary to expose the methods as it also creates partitions or something and then merges.

But I think this is useless for now.

I will make the methods private and have only Build as entrypoint, which creates the tmpdir at the start and removes it at the end so its more straigthforward. If we need to reuse the functions in other places, we can revisit that down the line.

But I really like that idea, will probably add it anyway

pkg/ops/rawDiskGeneration.go
Itxaka: This and the code below took part of whats left from my sanity. There has to be a better way of building freaking partition tables in 2024.

pkg/ops/rawDiskGeneration.go
Itxaka: yes, but where did the original value came from? I wonder if the default size we have for oem is indeed 64Mb xD

jimmykarily: That's why they call them "magic numbers" Itxaka. They are passed from wizard to wizard :D.

pkg/ops/rawDiskGeneration.go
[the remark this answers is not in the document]
Itxaka: added!

pkg/ops/rawDiskGeneration.go
Itxaka: ahh found it, itsd like that from the start https://github.com/kairos-io/kairos-agent/blob/main/pkg/constants/constants.go#L54

Also this shows that the efi partition should be 64 as well!

pkg/ops/rawDiskGeneration.go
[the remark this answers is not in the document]
Itxaka: done!

pkg/ops/rawDiskGeneration.go
Itxaka: done!

image-assets/azure.sh
```diff
@@ -1,17 +1,24 @@
 #!/bin/bash
 
+set -ex
 # Transform a raw image disk to azure vhd
```
jimmykarily, line 4: If the new go implementation works, shouldn't we delete this script altogether?

jimmykarily (approved):

image-assets/azure.sh
Itxaka: no this is a different script, this doesnt cover the transformation yet to gce/vhd

PR #147, opened 2024-12-23
Add tracking of time it takes each step to execute

Itxaka:
Useful to track how much do we take normally in the steps and improve on places that there may be a bottle neck or help us run things in parallel if possible.

Its set at trace level, which IIRC its not possible to set yet, so it needs a way of enabling trace level, maybe by env var or something?

### DIFF
--- a/deployer/steps.go
+++ b/deployer/steps.go
@@ -7,27 +7,48 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"time"
 
 	"github.com/kairos-io/AuroraBoot/pkg/ops"
 	"github.com/spectrocloud-labs/herd"
 )
 
+type callback func(context.Context) error
+
+// logStepTime logs the start and end time of a step
+// and the duration it took to complete
+func logStepTime(stepName string, stepFunc callback) callback {
+	return func(ctx context.Context) error {
+		start := time.Now()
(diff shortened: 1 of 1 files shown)


jimmykarily (approved):

deployer/steps.go
```diff
@@ -7,35 +7,56 @@ import (
...
+			internal.Log.Logger.Trace().Str("step", stepName).Time("start", start).Time("end", end).Str("duration", duration.String()).Msg("Completed step")
+		}
+		return err
+	}
+}
+
 func (d *Deployer) StepPrepNetbootDir() error {
 	return d.Add(opPrepareNetboot, herd.WithCallback(
-		func(ctx context.Context) error {
+		logStepTime(opPrepareNetboot, func(ctx context.Context) error {
```
jimmykarily, line 38: If we are going to do it in every case, maybe we can make it part of `Add` or `WithCallback`? I haven't checked the code to see but I assume it's possible to just call it once in one of those 2. If not, ignore me.

Itxaka: well I guess we could make it part of herd directly and have it as an option to track the step time and report it as part of the step completition or something, like the run or error values on each step

jimmykarily: If you think it's better...your call.

Itxaka:
hey look, Im stupid! https://github.com/spectrocloud-labs/herd/blob/master/dag.go#L34

PR #154, opened 2024-12-31
Raw bios image generation

Itxaka:
This will generate a BIOS raw image that behaves exactly the same as the EFI raw image.
Minimal image, able to truncate it to desired size.
Only recovery.
On first boot it will auto reset.
On second boot it will boot normally as an installed system.
If no config is provided a default one is given with at least the kairos/kairos user.


deployer/steps.go
```diff
@@ -126,16 +126,8 @@ func (d *Deployer) StepGenRawDisk() error {
...
-			herd.WithDeps(opDownloadISO), herd.WithDeps(opGenISO),
-		),
-		herd.IfElse(d.isoOption(),
-			herd.WithCallback(
-				ops.GenBIOSRawDisk(d.Config, d.isoFile(), d.rawDiskPath())),
-			herd.WithCallback(
-				ops.GenBIOSRawDisk(d.Config, d.isoFile(), d.rawDiskPath())),
-		),
-	)
+		d.imageOrSquashFS(),
```
Itxaka, line 129: Im not sure if this si correct. I.E. are we trying to create a BIOS image vs an EFI image or a MBR raw image VS a GPT raw image?

Notice that an MBR image can only support BIOS but an GPT image can support both BIOS and EFI. 

I think this was meant as BIOS image so the notation may be wrong or my approach is wrong :D

pkg/ops/rawDiskGeneration.go
```diff
@@ -415,6 +418,48 @@ func (r *RawImage) createEFIPartitionImage() (string, error) {
 	return efiPartitionImage.File, nil
 }
 
+// createBiosPartitionImage creates a BIOS partition image with grub artifacts and installs grub to it
+func (r *RawImage) createBiosPartitionImage() (string, error) {
+	// Would need to format it and then copy the grub files to it
+	// Also install grub against it
+	// Remember to use the grub artifacts from the rootfs to support secureboot
```
jimmykarily, line 425: what secure boot? This is for bios booting, no?

pkg/ops/rawDiskGeneration.go
```diff
@@ -415,6 +418,48 @@ func (r *RawImage) createEFIPartitionImage() (string, error) {
...
+	// Create a temp dir for copying the files to
+	tmpDirBios := filepath.Join(r.TempDir(), "bios")
+	err := fsutils.MkdirAll(r.config.Fs, tmpDirBios, 0755)
+	if err != nil {
+		internal.Log.Logger.Error().Err(err).Str("target", tmpDirBios).Msg("failed to create temp dir")
+		return "", err
+	}
+	defer r.config.Fs.RemoveAll(tmpDirBios)
```
jimmykarily, line 433: We use the whole `TempDir` to create the bios.img file from it, but we only cleanup the `bios` and `recovery-mout` directories from it (with `defer`). The final artifact is still in the `TempDir`. It feels somewhat weird because it's not clear if the `TempDir` has anything else on it. Is the `TempDir` what becomes the rootfs in the end? If yes, is it ok that we also use it to store build artifacts like the `bios.img` file below?

pkg/ops/rawDiskGeneration.go
```diff
@@ -758,3 +812,97 @@ func (r *RawImage) copyShimOrGrub(target, which string) error {
...
+	// boot.img goes to the first sector of the disk, sector 0 (512 bytes)
+	// core.img goes to the partition identified by bios_grub flag
+	// boot.img needs to have the LBA where core.img is located
+	// No idea where is that hardcoded
+	// core.img is not provided by grub packages, it has to be generated beforehand
+	args := []string{
+		"--target=i386-pc",
+		"--force", string(loopDevice),
+		fmt.Sprintf("--boot-directory=%s", tmpDirRecovery),
+	}
```
jimmykarily, line 893: I'm a bit confused. This command is not run in a chroot or anything. If I understand correctly, the code above will mount some more partitions from the img file, but we don't point this command to them in any way. If grub-install is only interested in the recovery partition, can't we run this command against that before we pack everything in an img file? I mean, while everything is still just a directory on the build machine. Then we wouldn't need to loop mount etc.

I'm sure I'm missing something :D

pkg/ops/rawDiskGeneration.go
Itxaka: indeed, not sure if this was a comment brougth from the efi partition image function which was duped or soomething?

pkg/ops/rawDiskGeneration.go
Itxaka: no, this is all wrong. Sorry about the confusing this seems to be a leftover from a thousand reworks. Bio.img is an empty diisk image, we dont need any of this mounts or anything, it just needs to be an empty image only with nothing inside as grub-install will then write core.img directly to that partition/

pkg/ops/rawDiskGeneration.go
Itxaka: well grub2-install does a lot of things behind the scenes.

we need the `tmpDirRecovery` which is the recovery partition mounted, as it will copy the proper artifcats into it (all the mod files and fonts and such)
we need to point to the full image as the install target `loopDevice` becuase it will install boot.img into sector 0 of the image and search for a partition in that device that matches the `bios_boot` flag and install the core.img to it.

So you need to have the final full image set up before running this command, you cant really do it before as all three things need to match for a proper grub install to it.

Is that clear?

jimmykarily: > Is that clear?

kind of. Let's say "clear enough" :D.

Itxaka:
kind of weird, this fails in remote but not locally.... the loop device being loop3 kind of shows that maybe something is leaving loop devices open behind the scenes... or other tests are hitting the loop device at the same time and there are collisions....

jimmykarily (approved):

PR #164, opened 2025-01-07
Improve raw image inputs and test

Itxaka:
- change disk.raw/disk.mbr into disk.efi/disk.bios
 - keep old values valid for those inputs
 - Expand test to cover BIOS raw image
 - Expand tests to cover more surface testing for expected values
 - Fix wrong swithc on deployer raw image generation


.github/workflows/tests.yml
```diff
@@ -117,10 +117,49 @@ jobs:
...
+          CPUS: 2
+          CREATE_VM: true
+          KVM: true
+          SECUREBOOT: false
         run: |
-          # TODO: permissions stuff should be fixed on aurora side, the output dir should contain only
-          # there required output image/iso and not the leftovers from the build process
-          sudo chmod -R 777 build
-          #earthly +test-bootable-raw
           export RAW_IMAGE=$(find $PWD/build -name *.raw)
```
jimmykarily, line 164: should also be changed to `.efi` ?

pkg/ops/rawDiskGeneration.go
```diff
@@ -853,7 +853,7 @@ func (r *RawImage) installGrubToDisk(image string) error {
 	// Get only the loop device without the /dev/ prefix
 	cleanLoopDevice := string(loopDevice)[5:]
 	recoveryLoop := fmt.Sprintf("/dev/mapper/%s%s", cleanLoopDevice, "p3")
-	err = unix.Mount(recoveryLoop, tmpDirRecovery, "ext2", 0, "")
+	err = unix.Mount(recoveryLoop, tmpDirRecovery, "ext4", 0, "")
```
jimmykarily, line 856: Why not?

```suggestion
	err = unix.Mount(recoveryLoop, tmpDirRecovery, agentConstants.LinuxFs, 0, "")
```

pkg/ops/disks.go
```diff
@@ -117,74 +117,6 @@ func GenArmDisk(src, dst string, do schema.Config) func(ctx context.Context) err
 	}
 }
 
-func GenBIOSRawDisk(config schema.Config, srcISO, dst string) func(ctx context.Context) error {
```
jimmykarily, line 120: Isn't this still used here? https://github.com/kairos-io/AuroraBoot/pull/164/files#diff-a40f78b4b8ba4fe8dfa30d4b7ed0b53546de415d98dba2dbcf5ad84099e3de07R131

.github/workflows/tests.yml
Itxaka: nope, this generates raw images without efi/bios in the name unfortunately. we cant do that without breaking our naming convention. Still bios images should not be the default

pkg/ops/disks.go
Itxaka: no, it was substituted for a new op with the same name

pkg/ops/rawDiskGeneration.go
```diff
@@ -142,7 +142,7 @@ func (r *RawImage) createOemPartitionImage(recoveryImagePath string) (string, er
 								FSLabel:    agentConstants.StateLabel,
 								Size:       uint(size),
 								PLabel:     agentConstants.StatePartName,
-								FileSystem: agentConstants.LinuxImgFs,
+								FileSystem: agentConstants.LinuxFs,
```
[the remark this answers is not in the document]
Itxaka: nope,  ext3 and 4 only bring improvements that are not beneficial in our case, due to the inmmutability of the images. Plus I think its the default in the agent as well, so if we decide to move from ext2 to ext3/4 for some reason, we only have once place to do it :D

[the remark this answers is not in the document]
jimmykarily: So we do we switch to ext4 here? ([LinuxFs is ext4](https://github.com/kairos-io/kairos-agent/blob/e6ed19177f9363b9786fb2ea357dc73a24fa9efa/pkg/constants/constants.go#L49))

pkg/ops/disks.go
jimmykarily: oh, `GenBiosRawDisk` vs `GenBIOSRawDisk` :D . I was sure I was looking at the same string.

.github/workflows/tests.yml
jimmykarily: Where is the old name expected? It would be better if we are consistent (intput flag matching the output file).

Itxaka: what old name? this generates a raw image before runnig the tests agaisnt it. Like we do with the isos, we want to minimize the changes if (when) we bump versions) so this is the easier way to pass tot he test what image to use for the raw test. I dont get what the old name refores to? we used to generate shitty disk.raw images if thats what you refer to? But that was terrible and already removed since the first changes to Aurora disk raw generatin, now it produces identical names as the isos and such but with a different extension (.raw .vhd.raw, etc)

pkg/ops/rawDiskGeneration.go
Itxaka: The resolution is, i clicked the wrong one loool

e2e/raw_bootable_test.go
```diff
@@ -38,11 +40,50 @@ var _ = Describe("raw bootable artifacts", Label("raw-bootable"), func() {
...
+		})
+
 		// This checks both that the disk is bootable and with secureboot enabled
-		By("Have secureboot enabled", func() {
-			output, err := vm.Sudo("dmesg | grep -i secure")
-			Expect(err).ToNot(HaveOccurred(), output)
-			Expect(output).To(ContainSubstring("Secure boot enabled"))
+		if os.Getenv("SECUREBOOT") == "true" {
+			By("Have secureboot enabled", func() {
+				output, err := vm.Sudo("dmesg | grep -i secure")
```
jimmykarily, line 59: Can we check for a bit more text than just "secure" ? `dmesg` produces lots of text. On my system for example:

```
~ $ sudo dmesg | grep secure
[79659.762609] intel-ipu6 0000:00:05.0: IPU6 in secure mode
[426107.139126] intel-ipu6 0000:00:05.0: IPU6 in secure mode
[513950.074595] intel-ipu6 0000:00:05.0: IPU6 in secure mode
[600414.944327] intel-ipu6 0000:00:05.0: IPU6 in secure mode
```
(probably nothing to do with secure boot)

jimmykarily (approved):
Only a nit. Looks good.

e2e/raw_bootable_test.go
Itxaka: ummm good point, I did not stole this from kairos so the implementation is worng.

Plus this will also match "secureboot disabled" LOL

Now watch me change it and it starting to fail....

Itxaka: ahh you rascal, almost got me!

This only gets the output that matches secure but then 2 lines undernmeath we match against the proper string!

```
Expect(output).To(ContainSubstring("Secure boot enabled"))
```

:D

PR #166, opened 2025-01-09
Support Rpi4 board raw image output and others

Itxaka:
no switch or anything, as long as the base image provided is a valid rpi image (identified by the KAIROS_MODEL entry) it generates an arm64 raw image that boots as expected on rpi4

Nothing special here, just the firmware files copied into the proper EFI dir so they are loaded on boot.

HINT: Maybe we could link to the upstream repo that provides those files and bundle them in the binary directly compressed and uncompress them on demand if we require them for rpi4? That would allow us to drop those luet packages and not having to deal with them, and if its a git repo, it should be easy to automate the bundling of the files.....

we would gain the ability to not depend on external packages for those and having a single source for them.

This also implements the finalizers for other boards but as I dont have any Im unable to test them. Im not even sure what those other boards use as a base image or rootfs?

This doesnt substitute the older methods, but creates a new path to generate the rpi4 (at least) images with a more standard approach and one that its used on other workflows. Generic raw image generation for x86, arm64 and boards is a nice thing to add.

Will need to update docs.


pkg/ops/rawDiskGeneration.go
```diff
@@ -231,17 +231,17 @@ func (r *RawImage) createRecoveryPartitionImage() (string, error) {
...
+	recoveryImage := &v1.Image{
 		File:       filepath.Join(tmpDirRecovery, "cOS", agentConstants.RecoveryImgFile),
 		FS:         agentConstants.LinuxImgFs,
 		Label:      agentConstants.SystemLabel,
 		Source:     v1.NewDirSrc(r.Source),
 		MountPoint: tmpDirRecoveryImage,
 	}
 	size, _ := config.GetSourceSize(r.config, recoveryImage.Source)
-	recoveryImage.Size = uint(size)
+	recoveryImage.Size = uint(size + 100)
```
mauromorales, line 242: why is it 100 here and 150 bellow? might be worth making them constants that tell the reason for the value

pkg/utils/fs.go
```diff
@@ -222,3 +222,66 @@ func CalcFileChecksum(fs v1.FS, fileName string) (string, error) {
 
 	return fmt.Sprintf("%x", h.Sum(nil)), nil
 }
+
+// CopyDir copies the contents of the source directory to the destination directory.
+func CopyDir(src string, dst string) error {
```
mauromorales, line 227: should this be part of the sdk?

pkg/ops/rawDiskGeneration.go
Itxaka: well 100 for the recovery.img and then for the partition which contains the recovery image it has to be a bit bigger as it cannot be the same size :D

pkg/utils/fs.go
Itxaka: or the standard golang library so people stops reimplementing it every week

pkg/ops/rawDiskGeneration.go
Itxaka: I mean, if the recovery image is 10Mb and then we also need to copy grub stuff or configs, the partition that has the image inside needs to be that size plus a bit more to account for those artifacts, it cannot be 10Mb exactly as just writing a single text file will make it not fit :D

pkg/utils/fs.go
mauromorales: LOL that's so true

pkg/ops/rawDiskGeneration.go
mauromorales: I'd add at least a comment with this info you mention, it's obvious after explained but I wouldn't have thought of it otherwise 😅

mauromorales (approved):

pkg/ops/rawDiskGeneration.go
Itxaka: added. Plus fixed the size of recovery as it needs to hold 2 images during upgrade, which I almost forgot!

PR #188, opened 2025-02-14
Remove uneedeed stuff

Itxaka:
- Scripts for boards and arm images are gone, we built in golang
 - orin images still have the script
 - remove deps not needed from the dockerfile
 - remove path for arm raw image build as we build on golang for all arches


mauromorales (approved):
That's pretty awesome!

Screenshot 2025-02-14 at 17 38 24

PR #189, opened 2025-02-21
Generate artifact name from rootfs values

Itxaka:
This was already done in the raw image but the iso defaulted to kairos.iso when we had all the info around in the rootfs

This patch just converges the output name for both artifacts into a single method so we produce constant names from a given rootfs values

Still allows to override the iso image name

### DIFF
--- a/internal/cmd/build-iso.go
+++ b/internal/cmd/build-iso.go
@@ -25,10 +25,10 @@ var BuildISOCmd = cli.Command{
 			Usage:   "The cloud config to embed in the ISO",
 		},
 		&cli.StringFlag{
-			Name:    "name",
+			Name:    "override-name",
 			Aliases: []string{"n"},
-			Value:   KairosDefaultArtifactName,
-			Usage:   "Basename of the generated ISO file",
+			Value:   "",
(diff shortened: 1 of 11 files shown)


internal/cmd/build-iso.go
```diff
@@ -27,10 +27,10 @@ var BuildISOCmd = cli.Command{
 			Usage:   "The cloud config to embed in the ISO",
 		},
 		&cli.StringFlag{
-			Name:    "name",
+			Name:    "override-name",
 			Aliases: []string{"n"},
-			Value:   KairosDefaultArtifactName,
```
jimmykarily, line 32: I suspect the constant is not needed anymore? (and can be deleted)

jimmykarily (approved):

internal/cmd/build-iso.go
Itxaka: its used somehwere else. We really need to rework this as soon as its possible :sob:

Itxaka:
ah this needs changes on the uki iso build side as well, oh my

jimmykarily (approved):

PR #190, opened 2025-02-21
Fix alpine RPI

Itxaka:


pkg/ops/rawDiskGeneration.go
```diff
@@ -741,6 +737,49 @@ func (r *RawImage) copyShimOrGrub(target, which string) error {
...
+				return fmt.Errorf("could not write file %s at dir %s from %s", "bootaa64.efi", target, filepath.Join("/efi", constants.EfiBootPath, "grub.efi"))
+			}
+			// Also copy it into the grub name to have the same files as in other flavors
+			err = utils.CopyFile(
+				r.config.Fs,
+				filepath.Join("/efi", constants.EfiBootPath, "grub.efi"),
+				filepath.Join(target, constants.EfiBootPath, "grub.efi"),
+			)
+			if err != nil {
+				return fmt.Errorf("could not write file %s at dir %s from %s", "grub.efi", target, filepath.Join("/efi", constants.EfiBootPath, "grub.efi"))
```
jimmykarily, line 759: better store this in a variable to avoid repetition. I mean the `filepath.Join("/efi", constants.EfiBootPath, "grub.efi"),` part. Maybe the others too (although they repeat less).

pkg/ops/rawDiskGeneration.go
```diff
@@ -741,6 +737,49 @@ func (r *RawImage) copyShimOrGrub(target, which string) error {
 		return fmt.Errorf("invalid which value: %s", which)
 	}
 
+	// Alpine does not provide a shim nor a grub file, so we need to copy the shim from the docker image that we ship as fallback artifcat
+	if model, flavor, err := r.GetModelAndFlavor(); flavor == "alpine" && model != "generic" && err == nil {
```
jimmykarily, line 741: this whole new block could be moved to a function `CopyAlpineShim`

.github/workflows/tests.yml
```diff
@@ -87,7 +87,7 @@ jobs:
           tags: auroraboot:latest
       - name: Run e2e tests
         run: |
-          sudo go run github.com/onsi/ginkgo/v2/ginkgo -v -r -p --covermode=atomic --coverprofile=coverage.out --timeout=2h --label-filter "!bootable && !raw-bootable" ./e2e
+          sudo go run github.com/onsi/ginkgo/v2/ginkgo -v -r -p --covermode=atomic --coverprofile=coverage.out --timeout=2h --label-filter "e2e" ./e2e
```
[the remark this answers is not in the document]
Itxaka: good catch!, adding it!

jimmykarily:
rpi 4 doesn't boot:

image

pkg/ops/rawDiskGeneration.go
```diff
@@ -979,3 +980,39 @@ func (r *RawImage) FinalizeImage(image string) error {
...
+		targetPath = filepath.Join(target, constants.EfiBootPath, "bootaa64.efi")
+	}
+
+	err = utils.CopyFile(
+		r.config.Fs,
+		grubPath,
+		targetPath,
+	)
+	if err != nil {
+		return fmt.Errorf("could not write file %s at dir %s from %s", "bootx64.efi", target, filepath.Join("/efi", constants.EfiBootPath, "grub.efi"))
```
jimmykarily, line 1003: This messages is misleading. `grub.efi` is not the source file you are using above (`bootx64.efi` or `bootaa64.efi`)

jimmykarily:
Now it gets further. Drops me the a grub shell.

Itxaka:
seems like the default grub efi that comes with the static/grub-efi does not hav e bundled ext support, so we cna use the livecd one instead

now it fails due to the kernel (invalid magic number) 

will try with a generic one

Itxaka:
generic arm64 fails as well.

Will check with amd64 to see if it also fails.

image-assets/prepare_nvidia_orin_images.sh
```diff
@@ -37,7 +37,15 @@ mkdir -p $WORKDIR/tmpefi
 
 # Create the EFI partition FAT16 and include the EFI image and a basic grub.cfg
 truncate -s $((20*1024*1024)) bootloader/efi.img
-cp -rfv /arm/raw/grubefi/* $WORKDIR/tmpefi
+mkdir -p $WORKDIR/tmpefi/EFI/BOOT
```
mudler, line 40: this needs QA on the orin by flashing

Itxaka: can you test this :D

Itxaka: basically the change is:

 - instead of using the package grub.cfg, we just add it directly here as its 4 lines (chainloader)
 - ignore the grub-efi package as its useless (contains the grub.efi for a given arch), everything is already in the grub-artifacts package
 - instead use the grub-artifacts package which has the proper directory hierarchy instead of having to copy 2 separated things into 2 different places.

also as a hint, the artifacts copied before (arm64-efi dir) for grub, were not used in that dir, so trying to load a module from that dir would still fail lol. They have to be under /grub2 dir as thats the prefix used.

You can verify that by booting this and trying to load a non existing module and it will show the path it expects the files to be at. That path is hardcoded when building the grub.efi so its not our choice.

A change to the artifacts package shouldbe done to move the arm64-efi/x86_64-efi modules and such into the /grub2 subdir so we can just dump the artifact package into the efi partition

mauromorales (approved):

mudler (approved):
Looking good!

PR #194, opened 2025-02-27
Improve image release

Itxaka:
Released images were being auto deleted after 6 hours due to the original images expiring
This makes it so we dont expire those and instead we clean them up after generating the final image


jimmykarily:
so the final image generation was taking more than 6 hours? I'm confused.

Itxaka:
> so the final image generation was taking more than 6 hours? I'm confused.

No, it looks like the label was still in the underlying images so when the manifest was created, it brought the expiration label with it, so the released multiarch image had an expiration of 6 hours :facepalm:

mauromorales (commented):
if we go with this approach I think it would be safe to remove this and just re-tag and push

https://github.com/kairos-io/kairos/commit/718088e3eba67f31fc8c2a475efba4bda76e494f

Itxaka:
> if we go with this approach I think it would be safe to remove this and just re-tag and push
> 
> [kairos-io/kairos@718088e](https://github.com/kairos-io/kairos/commit/718088e3eba67f31fc8c2a475efba4bda76e494f)

the problem with that is that it goes out of the way of using the docker build push action :(

Itxaka:
> if we go with this approach I think it would be safe to remove this and just re-tag and push
> 
> [kairos-io/kairos@718088e](https://github.com/kairos-io/kairos/commit/718088e3eba67f31fc8c2a475efba4bda76e494f)

Maybe the solution is jsut to push the images to the auroraboot repo anyways and leave them there, so you would have explicit VERSION-arm64 and VERSION-amd64 and then VERSION so they are there for usage anyway

Itxaka:
what about this @mauromorales ?

this would generate 

quay.io/kairos/auroraboot:v0.5.3-amd64
quay.io/kairos/auroraboot:v0.5.3-amd64
quay.io/kairos/auroraboot:v0.5.3

so if you pull quay.io/kairos/auroraboot:v0.5.3 you get whatever platform you are in, but you can still explicitely pull quay.io/kairos/auroraboot:v0.5.3-arm64 to build arm64 artifacts I guess?

mauromorales:
@Itxaka sorry for the delay, sure thing that lgtm

mauromorales (approved):

PR #197, opened 2025-02-28
add rpi5 support

zreigz:
This PR adds support for RPI 5. 
I've added a new plural luet repository to be able to test the build process.
I've also added the PR for packages: https://github.com/kairos-io/packages/pull/1293

You can test it with:
aurorabot image: `ghcr.io/pluralsh/aurora-boot:main`
container_image: `zreigz/alpine:3.19-standard-arm64-rpi5-v3.4.0-alpha14-k3s1.30.9-k3s1` which generally from RPI 4 but I have changed the model in `etc/kairos-release`

```
docker run --rm --privileged -v /var/run/docker.sock:/var/run/docker.sock -v $PWD/build/:/output --platform linux/arm64  ghcr.io/pluralsh/aurora-boot:main --set container_image=zreigz/alpine:3.19-standard-arm64-rpi5-v3.4.0-alpha14-k3s1.30.9-k3s1 --set state_dir=/output --set disable_http_server=true --set disable_netboot=true --set disk.raw=true
```


Itxaka:
nice work @zreigz ! Did you get the image to boot in an RPI5? 

The changes seem good, but the pacakge stuff, we usually get the files from a verified source and track updates and such so we migth need to rework that.

The idea would be to generate a single firmware package for RPIP tracking the https://github.com/raspberrypi/firmware/tree/master repo which provides dbt and overlay files for all rpi models, so we can easily build those with a single workflow.

zreigz:
Generally, I've used existing files from two distributions: 
 - dbt and overlays from [home-assistant](https://github.com/home-assistant)
 - u-boot from ultramarine linux 

We used the previous method to build the Kairos image (with build-arm-image.sh), so I combined all those files and prepared Dockerfile to create our custom AurorBoot Docker image. Then I was able to generate the working Kairos image for RPI 5.

This is our repo: https://github.com/pluralsh/edge/tree/main/rpi5

Thanks for the feedback! That approach makes sense.  I’ll try to work on this, but since I’m currently involved in other tasks, it might take some time.

Itxaka:
looks like uboot packages are coming from Fedora:

uboot-images-armv8 https://packages.fedoraproject.org/pkgs/uboot-tools/uboot-images-armv8/index.html

So those have rpi5 support then? Gonna have to have a look

Itxaka:
interesting, uboot ha now support for rpi5 but not for compute module or rpi 500 (whatever that is)

so a simple patch to extend those should be good enough, I dont see anything else of value in the fedora cutoms uboot build: https://src.fedoraproject.org/rpms/uboot-tools/blob/rawhide/f/rpi-Add-identifiers-for-the-new-RPi-5-series.patch

zreigz:
@Itxaka thanks for checking this rpi 5 support, let me know if I can help with some tests.

PR #210, opened 2025-03-07
Fix azure

jimmykarily:
Part of: https://github.com/kairos-io/kairos/issues/3198

### DIFF
--- a/pkg/ops/rawDiskOutput.go
+++ b/pkg/ops/rawDiskOutput.go
@@ -7,12 +7,13 @@ import (
 	"encoding/binary"
 	"encoding/hex"
 	"fmt"
-	"github.com/kairos-io/AuroraBoot/pkg/utils"
 	"io"
 	"io/fs"
 	"math"
 	"os"
 	"time"
+
+	"github.com/kairos-io/AuroraBoot/pkg/utils"
 
 	uuidPkg "github.com/gofrs/uuid"
 	"github.com/kairos-io/AuroraBoot/internal"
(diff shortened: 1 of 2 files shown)


pkg/ops/rawDiskOutput.go
```diff
@@ -44,8 +45,6 @@ func Raw2Azure(source string) (string, error) {
 	var finalSize int64
 	// Calculate the final size in bytes
 	finalSize = ((actualSize + constants.MB - 1) / constants.MB) * constants.MB
-	finalSize -= 512
```
Itxaka, line 47: So this is what its breaking it? what out because IIRC the checksum is calculated based on the size or something like that? I know I added it somehow for a checksum reason or something like that?

Well, if it works, it works :D

pkg/ops/rawDiskOutput.go
```diff
@@ -88,7 +87,7 @@ func Raw2Azure(source string) (string, error) {
 	}
 	sizeFile := fileInfo.Size()
 
-	if sizeFile%constants.MB != 0 {
+	if int64(size)%constants.MB != 0 {
```
Itxaka, line 90: size is the old size, but we are writing the header to the file afterwards, so the size has changed. Thats why it updates the stat before checking it, to get the actual real size on disk of it. Not sure if this is ok

jimmykarily: Azure wants the image size without the "footer" to be a multiple of MB so we don't need to subtract 512 (see the other change). This check becomes less important then, but I changed it anyway to make it work.

Itxaka (approved):
lets goooo

PR #211, opened 2025-03-12
[factory web ui] Basic features

mauromorales:
fixes https://github.com/kairos-io/kairos/issues/3251

https://github.com/user-attachments/assets/3b93eec6-bff1-4df6-81cb-164d516cd9f9


Dockerfile
```diff
@@ -17,8 +17,42 @@ RUN go build -ldflags "-X main.version=${VERSION}" -o auroraboot
...
+              xfsprogs \
+              xorriso \
+              zstd \
+              python3-cryptography python3-pefile # ukify deps
+
+# systemd-ukify systemd-boot
+# Install grub2-efi-x64 only on x86 arches
+RUN if [ "$(uname -m)" == "x86_64" ]; then dnf install -y grub2-efi-x64; fi
+# Install grub2-efi-arm64 only on arm64 arches
+RUN if [ "$(uname -m)" == "aarch64" ]; then dnf install -y grub2-efi-aa64; fi
```
jimmykarily, line 55: Does this work when building multi arch images? I mean, does `uname` report the correct thing?

internal/web/commands.go
```diff
@@ -0,0 +1,76 @@
+package web
+
+import (
+	"fmt"
+	"io"
+	"os/exec"
+)
+
+func buildRawDisk(containerImage, outputDir string) string {
```
Itxaka, line 9: cant this hook into the actual methods if aurora directly? I guess we lose the output that way?

internal/web/assets/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+ARG KAIROS_INIT_IMAGE=quay.io/kairos/kairos-init:${KAIROS_INIT_VERSION}
+
+FROM ${KAIROS_INIT_IMAGE} AS kairos-init
+
+FROM ${BASE_IMAGE} AS base
+ARG VARIANT=core
+ARG MODEL=generic
+ARG TRUSTED_BOOT=false
+ARG KUBERNETES_PROVIDER=k3s
+ARG KUBERNETES_VERSION=latest
```
jimmykarily, line 12: The kairos one has already diverged: https://github.com/kairos-io/kairos/blob/34c05ee585cb9adbd402f6c084e5a20d40bf9199/images/Dockerfile#L12

(accepts `VERSION` arg)

I wonder what is the canonical location for the dockerfile now. Since kairos repo is also using Auroraboot, maybe kairos should also use this one?

Dockerfile
```diff
@@ -17,8 +17,42 @@ RUN go build -ldflags "-X main.version=${VERSION}" -o auroraboot
...
+              rsync \
+              sbsigntools \
+              squashfs-tools \
+              sudo \
+              udev \
+              util-linux \
+              xfsprogs \
+              xorriso \
+              zstd \
+              python3-cryptography python3-pefile # ukify deps
```
Itxaka, line 49: this were dropped already, we dont use ukify anymore. Is this list updated?

Also dont install docker here if we are using the upstream one below :D

Dockerfile
```diff
@@ -17,8 +17,42 @@ RUN go build -ldflags "-X main.version=${VERSION}" -o auroraboot
...
+
+# systemd-ukify systemd-boot
+# Install grub2-efi-x64 only on x86 arches
+RUN if [ "$(uname -m)" == "x86_64" ]; then dnf install -y grub2-efi-x64; fi
```
Itxaka, line 53: this was dropped alreayd

go.mod
```diff
@@ -17,16 +17,19 @@ require (
...
 	github.com/kairos-io/kairos-sdk v0.7.3
 	github.com/kairos-io/netboot v0.0.0-20241104101831-1454e04fdb07
 	github.com/klauspost/compress v1.18.0
+	github.com/labstack/echo/v4 v4.13.3
 	github.com/mudler/go-processmanager v0.0.0-20240820160718-8b802d3ecf82
 	github.com/mudler/yip v1.15.0
 	github.com/onsi/ginkgo/v2 v2.23.0
 	github.com/onsi/gomega v1.36.2
 	github.com/otiai10/copy v1.14.1
+	github.com/robert-nix/ansihtml v1.0.1
```
jimmykarily, line 32: This project doesn't have many dependencies but hasn't got any commit for the last 3 years. We should keep an eye for possible vulnerabilities in any of the [deps](https://github.com/robert-nix/ansihtml/blob/master/go.sum).

go.mod
```diff
@@ -17,16 +17,19 @@ require (
 	github.com/gofrs/uuid v4.4.0+incompatible
 	github.com/google/go-containerregistry v0.20.3
 	github.com/hashicorp/go-multierror v1.1.1
+	github.com/joho/godotenv v1.5.1
```
Itxaka, line 20: is this new? I cant see anything using it?

internal/web/assets/Dockerfile
```diff
@@ -0,0 +1,18 @@
+ARG BASE_IMAGE=ubuntu:20.04
```
Itxaka, line 1: base image is 20.04 but we dont offer that on the webui. If this is an obligatory build-arg, should we remove the default?

internal/web/commands.go
Itxaka: Im worried that if we change something in the inputs we wont notice until this is broken :(

internal/web/commands.go
```diff
@@ -0,0 +1,76 @@
...
+	image string,
+	karosInitVersion string,
+	baseImage string,
+	variant string,
+	model string,
+	trustedBoot bool,
+	kubernetesProvider string,
+	kubernetesVersion string,
+) string {
+	return fmt.Sprintf(`docker build %s \
```
Itxaka, line 38: Maybe this is for the future, but cant we just hook into the containerd libs here and create the image in pure go directly?

internal/web/image.go
```diff
@@ -0,0 +1,21 @@
+package web
+
+import (
+	"os"
+	"path/filepath"
+)
+
+func prepareDockerfile(tempdir string) error {
+	// Copy the Dockerfile
+	dockerFile, err := assets.ReadFile("assets/Dockerfile")
```
Itxaka, line 10: we are reading to write it? Cant we have then the file directly here as pure text? and just write it? Or is it used somewhere else?

Itxaka:
looks good to me, my comments are mostly nits of for the future.

The only one is the Aurora Dockerfile which seems like its not updated as it has things that are not in the master Auroraboot Dockerfile. If thats fixes thats a thumbs up from me.

go.mod
mauromorales: used here https://github.com/kairos-io/AuroraBoot/blob/main/pkg/utils/common.go#L330

Dockerfile
mauromorales: aparently these are not needed anymore, and for finding the arch I think we actually use this: https://github.com/kairos-io/AuroraBoot/blob/main/pkg/utils/common.go#L334

internal/web/assets/Dockerfile
mauromorales: got rid of it for now

internal/web/assets/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+ARG KAIROS_INIT_VERSION=v0.2.3
```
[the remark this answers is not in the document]
mauromorales: bumped

internal/web/image.go
mauromorales: yup good idea, ended up doing it that way with a template

go.mod
mauromorales: created an issue for this kairos-io/kairos#3267

internal/web/commands.go
mauromorales: good point, created a ticket https://github.com/kairos-io/kairos/issues/3268

internal/web/app.go
```diff
@@ -0,0 +1,275 @@
...
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/fs"
+	"net/http"
+	"os"
+	"path/filepath"
+	"sync"
+
+	"github.com/google/uuid"
```
Itxaka, line 14: we are already bringing `github.com/gofrs/uuid` so adding a new dep for the same thing seems useless? Cant we reuse the existing uuid lib?

internal/web/image.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"bytes"
+	"os"
+	"path/filepath"
+	"text/template"
+)
+
+func prepareDockerfile(job JobData, tempdir string) error {
+	// Create a Dockerfile from a template
+	tmpl := `FROM quay.io/kairos/kairos-init:v0.3.0 AS kairos-init
```
Itxaka, line 12: very cool

Itxaka (approved):
look good enough for me rfor the basic thing already, really great job!

Itxaka:
oh, just though, we will need to add tests to this as well :sob:

internal/web/app.go
```diff
@@ -0,0 +1,275 @@
...
+		buf := make([]byte, 1024)
+		for {
+			n, err := r.Read(buf)
+			if err != nil {
+				if err != io.EOF {
+					pw.CloseWithError(err)
+				}
+				break
+			}
+			pw.Write(ansihtml.ConvertToHTML(buf[:n]))
```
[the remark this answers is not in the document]
mauromorales: i did a bit of investigating and yeah it's not part of the stdlib, but there are alternatives like https://github.com/muesli/termenv, will follow up in another ticket

internal/web/app.go
mauromorales: good call 💪

PR #222, opened 2025-03-18
Update diskfs dep

Itxaka:


jimmykarily:
I don't know what I'm looking at here. Mauro approved though so feel free to merge.

Itxaka:
this was failing because yip required and older diskfs version.

Will also probably fail as agent uses an old diskfs version as well. So agent needs update first

PR #230, opened 2025-03-25
Drop yip direct dep and add sentinel

Itxaka:
We were bringing yip just for a single file which we can easily generate with strings.

It also adds a sentinel check for the autoreset.
Creating a file at /oem/.autoreset.sentinel will prevent the autoreset from starting automatically so we can build raw images that do not autoreset


jimmykarily (approved):
nice!

Itxaka:
I tested with the following config:

```
#cloud-config

stages:
  initramfs:
    - commands:
      - touch /oem/.autoreset.sentinel
    - name: "Set user and password"
      users:
        kairos:
          passwd: "kairos"
          groups:
            - "admin"
    - name: "Set hostname"
      hostname: kairos-{{ trunc 4 .Random }}
```

And it behave correctyl as the reset was not triggered automatically. Im gonna change the sentinel to `.autoreset.skip` to make it clearer

PR #233, opened 2025-03-27
Use ext2 for all image types in raw image

Itxaka:
Seems like old grub versions can get very confused with ext4 capabilities and instead of ignoring those, they just died trying to recognize them.

Lets make this simple, we dont really need that much from ext4 in our case, so we can just move to use ext2 partitions and images instead.


mudler:
Agree, much more safer and for the same reasons we use ext2 in AGX Orin scripts

mudler (approved):

Issue #243, opened 2025-04-07
New Feature Request - Add Remote Redfish based virtual-media deployment


wrkode:
So to allow remote provisioning of hardware that which onboards a BMC (Board Management Controller), it would be necessary a method for Auroraboot to interact with the BMC Redfish API for power management and Kairos install image deployment.

This issues is being raised to track this need.

wrkode:
I've created for this PR https://github.com/kairos-io/AuroraBoot/pull/244

I'm sorry for the large PR.

wrkode:
PR#244 has been merged, which effectively closes this issue

PR #244, opened 2025-04-07
Add redfish remote install capability

wrkode:
This PR adds Redfish capability and is linked to issue https://github.com/kairos-io/AuroraBoot/issues/243
It is to be treated as experimental for the moment until extensive hardware testing can be conducted.
The Redfish implementations currently addressed are:

- HPE iLO
- SuperMicro Redfish v4
- Generic (default)
- DMTF (general DMTF to be used also for PiKVM)


examples/redfish/README.md
```diff
@@ -0,0 +1,49 @@
+# RedFish Deployment Examples
```
jimmykarily, line 1: This is similar to the information on the README. I suggest we keep only one of the 2 (or link to this page from the README and make the section there minimal).

pkg/ops/rawDiskGeneration_linux.go
```diff
@@ -0,0 +1,40 @@
+//go:build linux
+// +build linux
```
jimmykarily, line 2: wait... how is this change relevant?

pkg/redfish/client_test.go
```diff
@@ -0,0 +1,126 @@
...
+	defer server.Close()
+
+	// Create client
+	client, err := NewClient(server.URL, "user", "pass", true, 30*time.Second)
+	if err != nil {
+		t.Fatalf("Failed to create client: %v", err)
+	}
+
+	// Create a temporary ISO file for testing
+	tmpFile := "test.iso"
```
jimmykarily, line 111: Maybe use:
```
os.CreateTemp
```

pkg/redfish/vendor.go
```diff
@@ -0,0 +1,42 @@
...
+	// GetSystemInfo retrieves basic system information
+	GetSystemInfo() (*SystemInfo, error)
+	// DeployISO deploys an ISO image to the target system
+	DeployISO(isoPath string) (*DeploymentStatus, error)
+	// GetDeploymentStatus retrieves the current status of the deployment
+	GetDeploymentStatus() (*DeploymentStatus, error)
+}
+
+// NewVendorClient creates a new vendor-specific RedFish client
+func NewVendorClient(vendor VendorType, baseURL, username, password string, verifySSL bool, timeout time.Duration) (VendorClient, error) {
```
jimmykarily, line 29: Why are these vendors so different? Isn't this supposed to be a protocol they implement (redfish)?

jimmykarily (commented):
I did a scan of the code just to get an idea of what we are implementing here. I guess we'd need to test these somehow. If you have any idea on how these can be tested let us know (emulator? some service that exposes this protocol? other?)

pkg/redfish/vendor.go
wrkode: Unfortunately, vendors may (and do) change the redfish implementation according to their needs. Hence the different vendors needed. I'll start working on iDRAC (Dell), once and if the PR gets merged.

pkg/redfish/client_test.go
wrkode: good idea, I'll implement right away

go.mod
```diff
@@ -2,11 +2,11 @@ module github.com/kairos-io/AuroraBoot
 
 go 1.23.6
 
-toolchain go1.24.1
+toolchain go1.24.0
 
 // https://github.com/golang/go/blob/583d750fa119d504686c737be6a898994b674b69/src/crypto/x509/parser.go#L1014-L1018
 // For keys with negative serial number:
-godebug x509negativeserial=1
+// Note: x509negativeserial=1 is handled by the crypto/x509 package
```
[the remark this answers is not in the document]
wrkode: I can change the "note", but I'd like to know more as it shouldn't be like that.

wrkode:
I'm putting this PR back to draft to address the comments and resolve the conflicts

pkg/ops/rawDiskGeneration_linux.go
jimmykarily: No what I actually meant is why is this file needed at all. I can't find where the rest of this PR needs these functions.

go.mod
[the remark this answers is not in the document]
jimmykarily: From the [commit message](https://github.com/kairos-io/AuroraBoot/commit/cb118b39) :

```
* Allow certs with negative serial numbers

Not sure why the asus certs have a negative serial number but they do.
If the asus box had that others might too, so we should better support
it. The alternative would be to generate certs with positive serial
number for the tests.

https://github.com/golang/go/blob/master/src/crypto/x509/parser.go#L1014-L1018
https://github.com/microsoft/mssql-docker/issues/895#issuecomment-2327988940
```

I have and asus PN64 box here and we were using its built-in certs for some tests. And it turned out that those certs had a negative serial number. That's all I know.

pkg/ops/rawDiskGeneration_linux.go
wrkode: Oh totally I see what you mean.
I working on a CAPI provider for Auroraboot and that code is left over from it. cleaning up now

pkg/hardware/inspector.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+	}
+
+	return nil
+}
+
+// hasFeature checks if the system has a specific feature
+func (i *Inspector) hasFeature(info *SystemInfo, feature string) bool {
+	// Add feature detection logic here
+	// For now, we'll assume all systems have UEFI
```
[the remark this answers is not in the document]
wrkode: yes, Redifish supports and return such info

go.mod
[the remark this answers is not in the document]
wrkode: interesting. I never saw that. but definitely won't forget it :)

examples/redfish/README.md
wrkode: Consolidate into main README

jimmykarily (approved):

PR #253, opened 2025-04-14
Centralize the logger

Itxaka:
Also create a more generic logger for the efi builder so it can log into the same sink as Auroraboot


ci-robbot (approved):
Approved

internal/cmd/build-uki.go
```diff
@@ -28,7 +29,72 @@ import (
...
+}
+
+func (h *CustomHandler) WithGroup(name string) slog.Handler {
+	return &CustomHandler{l: h.l}
+}
+
+// Handle processes the log record and redirects it to your logger.
+func (h *CustomHandler) Handle(c context.Context, r slog.Record) error {
+	// Example: Redirect to your logger (replace this with your logger logic)
+	// Here, we just print the log message to the console.
```
jimmykarily, line 52:
```suggestion
```

internal/cmd/build-uki.go
```diff
@@ -28,7 +29,72 @@ import (
...
+	logMessage := r.Message
+	// Replace this with your logger's logic
+	// Process attributes
```
jimmykarily, line 55:
```suggestion
```

internal/cmd/build-uki.go
```diff
@@ -198,14 +264,16 @@ var BuildUKICmd = cli.Command{
 		if ctx.Bool("debug") {
 			logLevel = "debug"
 		}
-		logger := sdkTypes.NewKairosLogger("auroraboot", logLevel, false)
+		internal.Log = sdkTypes.NewKairosLogger("auroraboot", logLevel, false)
```
jimmykarily, line 267: Didn't happen in this PR, but I find it weird that we defined a package named "internal" (given "internal" has a special meaning in Go). We should rather create a subpackage inside internal for the logger.

jimmykarily (approved):

internal/cmd/build-uki.go
Itxaka: Well, isn't internal meant for things that are not going to be reused by other things? Like private methods in other languages and such ? That's what I understood by the internal package in go, things that you don't want to expose to others when they import your PKG but that you want to use internally.

I think we do the same in a couple of other packages in which we set a exported logger for our app to use from anywhere (so kind of like a local global var) but put it in internal so when this package is imported, they can't use that internal.Log instance.


I dunno, maybe a public PKG global logger makes more sense so it can be override easily by someone calling it?

In not that happy with this, but it seems to work without having to explicitly passing the logger instance everywhere which makes the code nicer IMO for self contained apps. Otherwise is back to pass the logger around or having a config object with the logger inside or something uglier.

Thoughts?

Itxaka: I see, can do

PR #264, opened 2025-05-02
Fix web UI builds

jimmykarily:
The builds where broken since we stopped shelling out to the same binary: https://github.com/kairos-io/AuroraBoot/pull/258

because the DAG we build didn't have the step to copy the cloud config in place. I tried to "cherry-pick" only the relevant steps but I can't get a working DAG. I decided to just call RegisterAll and disable what's possible using conditions like `DisableNetboot`.

### DIFF
--- a/internal/web/commands.go
+++ b/internal/web/commands.go
@@ -48,21 +48,18 @@ func buildRawDisk(containerImage, outputDir string, ws io.Writer) error {
 		DisableNetboot:    true,
 	}
 
-	// Create the deployer
+	// Create the deployer with proper initialization
 	d := deployer.NewDeployer(config, artifact, herd.EnableInit)
 
-	// Register the necessary steps
-	for _, step := range []func() error{
-		d.StepPrepNetbootDir,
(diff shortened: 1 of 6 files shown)


internal/web/commands.go
```diff
@@ -90,32 +88,32 @@ func buildISO(containerImage, outputDir, artifactName string, ws io.Writer) erro
...
 	}
 
-	// Create the config
-	config := schema.Config{
-		State: outputDir,
-		ISO: schema.ISO{
-			OverrideName: artifactName,
-		},
+	// Read the config using the shared config package
+	config, _, err := config.ReadConfig("", "", nil)
```
jimmykarily, line 92: Initially I thought the reason it as failing was because I failed to set the cloud config. I tried to build the config using `ReadConfig` and this is the reason I extracted it to a separate package in this PR. I still think it's better to re-use this function but I can revert this change if if feels irrelevant.

mauromorales (approved):

PR #274, opened 2025-05-12
3308 factory api

jimmykarily:
Fixes https://github.com/kairos-io/kairos/issues/3308

This PR:
- Introduces the concept of "workers". By passing the `--create-worker` flag to the `web` command, you get a worker spawn up as a go routine, which gives you the same functionality as before (no need to start any workers manually). If the user wants, they can spin up more workers using the new `worker` command.
- Introduces an API that implements the whole build flow, allows the user to query the server for a specific job's artifacts, implements endpoints to check the job status, logs etc.
- Makes the old web UI rely on the worker for the job building making the endpoint only stream logs.
- Implements a filesystem-based storage for jobs. This way if the server process dies, when it get's up, it will still has access to the old jobs.
- Stores multiple job artifacts in a way that they can be queried again. This will allow us to show old jobs in the Web UI allowing the user to fetch the artifacts again if needed.
- Stores jobs logs in the filesystem as well, allowing the user to fetch them at any time.
- Fixes some issues with logs not showing in the web UI because auroraboot was using new loggers for some methods.

What this PR doesn't do:

- Doesn't do any authentication on the workers. The workers are trusted to be the worker they say they are. We will need to introduce tokens or some other mechanism to make sure only the allowed processes can talk to the server.
- Doesn't change the web UI to fully leverage the API. We still have a different endpoint to register a new job for example. The API allows all the actions so that the Web UI can be a single page application fully in JS.
- Doesn't show old jobs or other jobs currently running. This is possible but let's wait for [the frontend framework research](https://github.com/kairos-io/kairos/issues/3400) before we jump into more complicated features in the front end.


internal/worker/worker.go
```diff
@@ -0,0 +1,134 @@
...
+			continue
+		}
+		defer ws.Close()
+
+		// Process the job
+		logMessage := fmt.Sprintf("Processing job %s:\n", job.JobID)
+		logMessage += fmt.Sprintf("  Variant: %s\n", job.Job.JobData.Variant)
+		logMessage += fmt.Sprintf("  Model: %s\n", job.Job.JobData.Model)
+		logMessage += fmt.Sprintf("  Image: %s\n", job.Job.JobData.Image)
+		logMessage += fmt.Sprintf("  Version: %s\n", job.Job.JobData.Version)
```
jimmykarily, line 65: This is were the actual building will happen (TODO)

internal/web/api_handlers.go
```diff
@@ -0,0 +1,355 @@
...
+
+// HandleQueueBuild creates a new build job and adds it to the queue
+func HandleQueueBuild(c echo.Context) error {
+	var req jobstorage.JobData
+	if err := c.Bind(&req); err != nil {
+		return c.JSON(http.StatusBadRequest, map[string]string{"error": "Invalid request body"})
+	}
+
+	// Validate required fields
+	if req.Variant == "" || req.Model == "" || req.Image == "" {
```
jimmykarily, line 30: Are these the only ones required? Are they enough for a build?

internal/web/api_handlers.go
```diff
@@ -0,0 +1,275 @@
...
+	jobID := c.Param("job_id")
+
+	// Verify the job exists
+	if _, exists := jobsData[jobID]; !exists {
+		return c.JSON(http.StatusNotFound, map[string]string{"error": "Job not found"})
+	}
+
+	// Open the log file in read-only mode
+	logFile := filepath.Join(logsDir, jobID+".log")
+	file, err := os.OpenFile(logFile, os.O_RDONLY, 0644)
```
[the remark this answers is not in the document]
jimmykarily: The job ID is sanitized in the `jobstorage.GetJobPath` function which is used by all the other functions.

jimmykarily:
build-image test is failing because of this: https://status.redhat.com/incidents/k7kvfvgfrbdf
but since I don't expect that to break with my changes, I'm opening this for review.

internal/web/api_handlers.go
Itxaka: version, registry and base image

the other are set by default values in init itself so we should only need those threee to generate a Kairos image.

internal/cmd/web.go
```diff
@@ -22,9 +24,33 @@ var WebCMD = cli.Command{
...
 	},
 	Action: func(c *cli.Context) error {
 		os.MkdirAll(c.String("artifact-dir"), os.ModePerm)
-		return web.App(c.String("address"), c.String("artifact-dir"))
+		os.MkdirAll(c.String("builds-dir"), os.ModePerm)
+
+		// If create-worker flag is set, start a worker in a goroutine
+		if c.Bool("create-worker") {
+			workerID := "local-worker"
+			w := worker.NewWorker("http://localhost"+c.String("address"), workerID)
```
mauromorales, line 45: can we rename address to port?

internal/web/app.go
```diff
@@ -39,24 +44,25 @@ func getFileSystem(useOS bool) http.FileSystem {
...
-	Model                  string `json:"model"`
-	TrustedBoot            bool   `json:"trusted_boot"`
-	KubernetesDistribution string `json:"kubernetes_distribution"`
-	KubernetesVersion      string `json:"kubernetes_version"`
-	Image                  string `json:"image"`
-	Version                string `json:"version"`
-}
-
-func App(listenAddr, outDir string) error {
+func App(listenAddr, outDir, buildsDirectory string, config ...AppConfig) error {
```
mauromorales, line 47: Would it make sense to put everything in the AppConfig instead of different parameters?

internal/web/app/index.js
```diff
@@ -150,17 +150,17 @@
...
           const message = event.data;
+          if (!message.trim()) return; // Skip empty messages
+
           updateStatus(message);
-          outputElement.innerHTML += `${convert.toHtml(message)}\n`;
+          // Split message by newlines and wrap each line in a div
+          const lines = message.split('\n');
+          for (const line of lines) {
+            if (line.trim()) { // Only add non-empty lines
+              outputElement.innerHTML += `<div>${convert.toHtml(line)}</div>`;
```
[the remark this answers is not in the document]
jimmykarily: fixed

internal/cmd/web.go
jimmykarily: I don't mind changing it. But it comes from the command's `Flags` above and it was called `address` so far. Since we don't use the web interface in CI or anything like that, I don't expect anything to break from this change.

internal/web/app.go
jimmykarily: Although I hate "config" arguments, in this one I think it makes sense. As long as it's only used for App initialization.

jimmykarily: done

internal/web/api_handlers.go
jimmykarily: We don't pass the registry to kairos-init (yet): https://github.com/kairos-io/AuroraBoot/blob/130dac6cf787df3a77caf0cb498077d070c9333c/internal/web/image.go#L17

I guess it's because we need to bump kairos-init to a newer version (some lines before).

Having said that, the Web UI doesn't let the user specify the registry either, we would need to add a field for that. Looking at kairos-init, it only requires the registry so that it can write it here: https://github.com/kairos-io/kairos-init/blob/4dbb9be2b224324b9d13e19b4f242d599d08a9d0/pkg/stages/steps_init.go#L214

In other words, the only thing that would break if the registry was missing, would be the `kairos-agent upgrade list-releases` :facepalm: . We pay that command dearly.
For people who intend to upgrade to images that have been built or tested in CI or at found by different means, forcing them to specify the registry makes little sense. They will probably never use `list-releases`.

A possible solution is to make `list-releases` complain when the variable is not set in `/etc/kairos-release` and allow them to pass it to the `list-releases` command using a flag. This way, everyone can say "Hey kairos-agent, tell me which releases are available in this registry (flag specified) using the rest of the values in my kairos-release file". We can make the registry optional in kairos-init this way and the only people that care are those that don't set it and try to run `list-releases`. In that case, they can just pass it at runtime.

@Itxaka what do you think?

jimmykarily: Made Version and Image mandatory and left the registry out since it's not required in the referenced kairos-init version. We can add it when we decide on the above.

Itxaka: I think that if we can do that and drop the registry flag or make it optional it would be amazing, and moving it to the list releases is a very good fix for this!

internal/cmd/web.go
mauromorales: if we don't have much deps on this yet I'd say let's do but don't let that block merging

jimmykarily: Actually, the "address" field, is propagated down to: https://go.dev/src/net/http/server.go#L2959

which is indeed the full address. It is something like `host:port` with `:port` also being valid. I think the right fix is to change the code here, split in `:` and take the second part (the port).

jimmykarily: Fixed.

internal/web/api_handlers.go
jimmykarily: Created a new ticket for this: https://github.com/kairos-io/kairos/issues/3402

mauromorales (approved):

PR #276, opened 2025-05-14
Whatever

Itxaka:


pkg/utils/pxeUki.go
```diff
@@ -0,0 +1,315 @@
...
+		dhcp.Close()
+		return err
+	}
+	pxe, err := net.ListenPacket("udp4", fmt.Sprintf("%s:%d", "0.0.0.0", nbConstants.PortPXE))
+	if err != nil {
+		dhcp.Close()
+		tftp.Close()
+		return err
+	}
+	// 5 buffer slots, one for each goroutine, plus one for
```
jimmykarily, line 41: I see only 4 goroutines and no Shutdown() ?

pkg/constants/constants.go
```diff
@@ -37,6 +37,12 @@ var BootHybrid []byte
 //go:embed grub_live_bios.cfg
 var GrubLiveBiosCfg []byte
 
+// EfiKeyEnroller is the key enroller for the efi image which gets the keys from the ipxe server and enrolls them
+// source at https://github.com/Itxaka/kairos-efi-booter/
```
mudler, line 41:
```suggestion
// source at https://github.com/kairos-io/efi-key-enroller/
```

pkg/utils/pxeUki.go
```diff
@@ -0,0 +1,306 @@
...
+			http.ServeFile(w, r, filePath)
+			return
+		}
+		log.Logger.Info().Str("url", r.URL.Path).Msg("File not found (case-insensitive)")
+		w.WriteHeader(http.StatusNotFound)
+		w.Write([]byte("404 not found"))
+	})
+
+	http.Handle("/", handler)
+	listenAddr := os.Getenv("HTTP_LISTEN_ADDR")
```
mudler, line 296: maybe better to have this as a cmd flag? wdyt?

mudler (approved):
just small nit, didn't tested but changes looks good

pkg/utils/pxeUki.go
Itxaka: Umm this was added by copilot but it's wrong, http server is harcoded to run in there I think as the whole thing doesn't deal with different http ports at all. So I think this need to be dropped

PR #277, opened 2025-05-15
3403 api docs

jimmykarily:
Part of: https://github.com/kairos-io/kairos/issues/3403


internal/web/app.go
```diff
@@ -22,10 +22,6 @@ import (
 var staticFiles embed.FS
 
 var mu sync.Mutex
-var artifactDir string
-
-//go:embed assets
-var assets embed.FS
```
jimmykarily, line 28: these were not used anywhere

internal/web/app/index.js
```diff
@@ -156,7 +154,9 @@ document.addEventListener('DOMContentLoaded', () => {
           updateStatus(message);
           // Create a pre element to preserve whitespace and prevent wrapping
           const pre = document.createElement('pre');
-          pre.textContent = convert.toHtml(message);
+          // Strip ANSI color codes using regex
+          const strippedMessage = message.replace(/\x1b\[[0-9;]*m/g, '');
+          pre.textContent = strippedMessage;
```
jimmykarily, line 159: We were wrapping the message in `<pre>`. Replacing color codes with <span> makes no sense because they are rendered as literal strings. We simply remove the color codes here.

internal/worker/worker.go
```diff
@@ -381,7 +378,7 @@ func (w *Worker) updateJobStatus(jobID string, status jobstorage.JobStatus) erro
 
 func prepareDockerfile(job jobstorage.JobData, tempdir string) error {
 	// Create a Dockerfile from a template
-	tmpl := `FROM quay.io/kairos/kairos-init:v0.3.0 AS kairos-init
+	tmpl := `FROM quay.io/kairos/kairos-init:v0.4.3 AS kairos-init
```
jimmykarily, line 381: This version doesn't make the registry yet required (the next release will).

internal/worker/worker.go
```diff
@@ -411,7 +408,7 @@ RUN rm /kairos-init`
 }
 
 func buildOCI(contextDir, image string) string {
-	return fmt.Sprintf(`docker build %s -t %s`, contextDir, image)
+	return fmt.Sprintf(`docker buildx build --load --progress=plain %s -t %s`, contextDir, image)
```
jimmykarily, line 411: We were seeing some deprecation warning because of using `docker build`. This fixes it. `--progress=plain` is to print the logs without manipulating the terminal be replacing lines. `--load` means load the image to the docker daemon.

.github/workflows/tests.yml
```diff
@@ -9,7 +9,7 @@ concurrency:
   cancel-in-progress: true
 jobs:
   unit-tests:
-    runs-on: kvm
+    runs-on: ubuntu-latest
```
jimmykarily, line 12: The docker version on our workers was too old for `buildx`

ci-robbot (approved):
The pull request adds Swagger documentation and Redoc integration, which improves API visibility and usability. The CI checks are all passing, and the changes are well-structured. I approve this PR.

internal/web/app/swagger.json
```diff
@@ -0,0 +1,644 @@
+{
```
mauromorales, line 1: does this file need to be included in the commit? or will it be updated at runtime? it's been a while since I played with swagger

mauromorales (approved):

internal/web/app/swagger.json
jimmykarily: It is served as a static file by the web server. So it has to be there. The Dockerfile builds it every time we build the image.

PR #278, opened 2025-05-15
Improve logs

jimmykarily:
- Don't polute the logs with the prefix
- Print the job id and the worker once in the beginning (after we acquired a websocket connection)
- Show a horizontal scroll bar instead of wrapping long text
- fix problem where we close the connection before we get all the logs
- fix logos (Fedora was showing Ubuntu and other issues). This is what they look like now:

image


internal/web/app.go
```diff
@@ -225,7 +225,25 @@ func webSocketHandler(c echo.Context) error {
 		// Create a buffer for reading
 		buf := make([]byte, 1024)
 		for {
```
jimmykarily, line 227: this block was reading 1024 bytes of data and if the job was completed, it just stopped on the next iteration. Now it reads until there is no new data and then checks if the job is complete. This was if a job completes before we get to read all the data, we still print them all.

internal/web/app/index.html
```diff
@@ -505,7 +505,7 @@ <h3 class="text-xl font-semibold text-gray-900 dark:text-white">
                                                 <div class="relative w-11 h-6 bg-gray-200 peer-focus:outline-none peer-focus:ring-4 peer-focus:ring-blue-300 dark:peer-focus:ring-blue-800 rounded-full peer dark:bg-gray-400 peer-checked:after:translate-x-full rtl:peer-checked:after:-translate-x-full peer-checked:after:border-white after:content-[''] after:absolute after:top-[2px] after:start-[2px] after:bg-white after:border-gray-300 after:border after:rounded-full after:h-5 after:w-5 after:transition-all dark:border-gray-600 peer-checked:bg-blue-600 dark:peer-checked:bg-blue-600"></div>
                                                 <span class="ms-3 text-sm font-medium text-gray-900 dark:text-gray-300">Show Logs</span>
                                             </label>
-                                            <div id="output" class="bg-gray-900 text-green-400 p-6 rounded-lg h-64 overflow-y-auto font-mono text-sm whitespace-pre-wrap mt-4 hidden">
+                                            <div id="output" class="bg-gray-900 text-green-400 p-6 rounded-lg h-64 overflow-y-auto overflow-x-auto font-mono text-sm whitespace-pre mt-4 hidden">
```
jimmykarily, line 508: adds a horizontal scrollbar when lines are too long

internal/web/app/index.js
```diff
@@ -154,13 +154,10 @@ document.addEventListener('DOMContentLoaded', () => {
...
           updateStatus(message);
-          // Split message by newlines and wrap each line in a div
-          const lines = message.split('\n');
-          for (const line of lines) {
-            const div = document.createElement('div');
-            div.textContent = convert.toHtml(line);
-            outputElement.appendChild(div);
-          }
+          // Create a pre element to preserve whitespace and prevent wrapping
+          const pre = document.createElement('pre');
```
jimmykarily, line 158: the div was still wrapping the text, thus the scrollbar didn't appear. `pre` fixes it.

internal/worker/worker.go
```diff
@@ -64,43 +74,63 @@ func (m *MultiWriterWithWebsocket) Write(p []byte) (n int, err error) {
...
 			return n, err
 		}
 
-		// Send to the websocket
-		if err := websocket.Message.Send(m.ws, message); err != nil {
-			// We've already written to the regular writer, so we can't return 0
-			// Instead, we'll return the number of bytes written to the regular writer
-			return n, err
+		// Send to the websocket with retry if available
+		if m.ws != nil {
```
jimmykarily, line 88: we only send over the wire, if we have a connection established. Initially the worker doesn't yet have a ws connection. But we still use the same writer everywhere.

internal/web/api_handlers.go
```diff
@@ -179,20 +179,38 @@ func HandleGetBuildLogs(c echo.Context) error {
...
+			// Check job status after reading all currently available data
+			job, err := jobstorage.ReadJob(jobID)
+			if err != nil {
+				websocket.Message.Send(ws, fmt.Sprintf("Error reading job status: %v", err))
+				return
+			}
+
+			// If job is complete or failed, close the connection
+			if job.Status == jobstorage.JobStatusComplete || job.Status == jobstorage.JobStatusFailed {
+				websocket.Message.Send(ws, "Job complete, closing connection.")
```
mauromorales, line 209: Maybe the message should not be "Job complete, if the status is failed?

mauromorales (approved):

internal/web/api_handlers.go
jimmykarily: It was more in the sense of "finished" (reaching a terminal state). I fixed it to report the actual terminal state.

PR #302, opened 2025-06-06
Allow users to choose which artifacts to download

mauromorales:
relates to https://github.com/kairos-io/kairos/issues/3266


jimmykarily:
I suspect something is not proper in the HTML/css:

image


The bottom border doesn't show up and there is a little gap in the border between the "artifacts" box and the "version" box above it.

PR #309, opened 2025-06-11
Use icons for the artifacts

mauromorales:
Once we add the cloud images, this section will be a bit crowded, while I don't expect someone to choose all the values, it's still better to just have the images on the header and not the full descriptions


Itxaka (approved):
Fancy!

PR #312, opened 2025-06-17
Several fixes to Aurora

Itxaka:
As we can have different iso names for some of the steps, we should try to get the most recent iso based on some basic parameters:

 - default to kairos.iso so we maintain backwards compatibility
 - otherwise iterate over the dest dir to pick up whatever iso image was generated

this seems to work as expected but has the side effect of moving the iso argument to be a function so it can be executed when the step is runnig instead of being run at the start of the deployment, where the image may not be available yet.

This fixed netbooting from an iso, which is built from a container before netboot starts

Fixes (and things to test before approving):

 - Downloading from a released ISO and serving netboot artifacts: 
 ```
docker run -v "$PWD"/config.yaml:/config.yaml --rm -ti --net host aurora \
        --set "artifact_version=v3.4.2" --set "release_version=v3.4.2" \
        --set "flavor=debian" --set "flavor_release=12" \
        --set repository="kairos-io/kairos" --debug
```
 - Building an iso from a docker image and serving netboot artifacts from it:
```
docker run --rm -ti -v /var/run/docker.sock:/var/run/docker.sock --net host aurora --debug --set "container_image=ubuntu-init:24.04"
```
 - Not passing the state_dir on some calls, now it will default to a decent dir and create it beforehand
 - Raw image building. Not sure what the hell we were thinking having 2 different paths for sources? Lets keep this straigth until we can rework this but seems simple enough, raw images are generated from container sources, its like that on the docs and generating them from isos is not documented anywhere so lets drop that and keep a single straigth way of building them until we can rework the entrypoints, otherwise everything is flimsy.

Things changed:

 - Reorder some stuff, like state dir should come next os netboot and rootfs dirs and created on top of it
 - Drop some uneeded steps and merged them with existing one. Extract artifacts or download artifacts are now converged
 - Add some deps between steps, like inject cloud config for downloaded+generated isos
 - MOVE EVERYTHING RELATED TO PATHS INTO CALLBACKS. This one was biting us. We resolved all paths on the deploy step, when we generated the DAG but that didnt work as expected. Some stuff is not there as it depends on previous steps, so we have to call the path resolve inside the callback so they are done at that exact moment, not beforehand. This solved a lot of issues with passing around names and paths and artifacts.

Fixes https://github.com/kairos-io/kairos/issues/3475
Fixes https://github.com/kairos-io/kairos/issues/3425

### DIFF
--- a/deployer/register.go
+++ b/deployer/register.go
@@ -19,7 +19,6 @@ func RegisterAll(d *Deployer) error {
 		d.StepDownloadISO,
 		d.StepExtractNetboot,
 		// Ops to generate RAW disk images
-		d.StepExtractSquashFS,
 		d.StepGenRawDisk,
 		d.StepGenMBRRawDisk,
 		d.StepConvertGCE,
--- a/deployer/register.go
+++ b/deployer/register.go
@@ -10,19 +10,14 @@ import (
 // This registers all steps for the top level Auroraboot command.
 func RegisterAll(d *Deployer) error {
 	for _, step := range []func() error{
+		d.StepPrepDestination,
 		d.StepPrepTmpRootDir,
 		d.StepPrepNetbootDir,
-		d.StepPrepISODir,
 		d.StepCopyCloudConfig,
 		d.StepDumpSource,
 		d.StepGenISO,
-		d.StepExtractNetboot,
-		//TODO: add Validate step
-		// Ops to download from releases
-		d.StepDownloadInitrd,
-		d.StepDownloadKernel,
-		d.StepDownloadSquashFS,
 		d.StepDownloadISO,
+		d.StepExtractNetboot,
 		// Ops to generate RAW disk images
 		d.StepExtractSquashFS,
 		d.StepGenRawDisk,
--- a/deployer/steps.go
+++ b/deployer/steps.go
@@ -2,6 +2,7 @@ package deployer
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"path/filepath"
 	"strconv"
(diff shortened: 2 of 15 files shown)

### CODE REVIEW

**Comment by Itxaka** [comment]
this does not fix the netboot part (i.e. pass an artifact/release, download the iso and extract the netboot artifacts)

For some reason, it tries to download all the artifacts, while it should extract them from the iso instead

**Comment by Itxaka** [comment]
This now fixes netboot as well

```
$ docker run -v "$PWD"/config.yaml:/config.yaml --rm -ti --net host aurora \
        --set "artifact_version=v3.4.2" --set "release_version=v3.4.2" \
        --set "flavor=debian" --set "flavor_release=12" \
        --set repository="kairos-io/kairos" --debug
```

File: deployer/register.go [hunk 1]
```diff
@@ -10,21 +10,15 @@ import (
...
 		d.StepPrepTmpRootDir,
 		d.StepPrepNetbootDir,
-		d.StepPrepISODir,
 		d.StepCopyCloudConfig,
 		d.StepDumpSource,
 		d.StepGenISO,
-		d.StepExtractNetboot,
-		//TODO: add Validate step
-		// Ops to download from releases
-		d.StepDownloadInitrd,
```
Itxaka, line 22: This is no longer supported, we dont publish netboot artifacts

internal/cmd/app.go
```diff
@@ -86,19 +86,26 @@ func GetApp(version string) *cli.App {
 				return err
 			}
 
+			if d.Config.State == "" {
```
Itxaka, line 89: being safe, as we have several entrypoints, we have to make sure to set the default value for this if nothing is passed

internal/cmd/app.go
```diff
@@ -86,19 +86,26 @@ func GetApp(version string) *cli.App {
...
-			if err != nil {
-				internal.Log.Logger.Error().Err(err).Msg("Failed to clean up tmp root dir")
+			if errCleanup != nil {
 				// Append the cleanup error to the main errors if any
 				err = multierror.Append(err, errCleanup)
 			}
 
+			// If there are errors, write the DAG to help debugging
+			if err != nil {
+				d.WriteDag()
```
Itxaka, line 107: this helps A LOT as it shows where the errors come from and steps and such!

internal/cmd/app.go
```diff
@@ -86,19 +86,26 @@ func GetApp(version string) *cli.App {
...
 			d.WriteDag()
 			if err := d.Run(ctx.Context); err != nil {
 				return err
 			}
 
 			err = d.CollectErrors()
 			errCleanup := d.CleanTmpDirs()
-			if err != nil {
-				internal.Log.Logger.Error().Err(err).Msg("Failed to clean up tmp root dir")
+			if errCleanup != nil {
```
Itxaka, line 100: this stupid mistake made it alwasy fail with the same errors xD

internal/cmd/build-iso.go
```diff
@@ -105,11 +105,15 @@ var BuildISOCmd = cli.Command{
...
+		if c.State == "" {
+			c.State = "/tmp/auroraboot"
+		}
+
 		d := deployer.NewDeployer(c, r, herd.EnableInit)
 		for _, step := range []func() error{
+			d.StepPrepDestination,
 			d.StepPrepNetbootDir,
 			d.StepPrepTmpRootDir,
-			d.StepPrepISODir,
```
Itxaka, line 112: renamed, as its not really the isodir

pkg/ops/iso.go
```diff
@@ -199,7 +202,13 @@ func InjectISO(dst, isoFile string, i schema.ISO) func(ctx context.Context) erro
 		if err != nil {
 			return err
 		}
-		internal.Log.Logger.Info().Msgf("Wrote '%s'", injectedIso)
+		os.Remove(isoFile)                    // remove the original iso file
```
Itxaka, line 205: instead of keeping 2 isos, we keep only one the injected one, if you want a clean iso, dont pass a cloud config, solved, wtf are you even doing bro?

deployer/steps.go
```diff
@@ -52,17 +59,31 @@ func (d *Deployer) CleanTmpDirs() error {
...
 	}))
 }
 
 func (d *Deployer) StepCopyCloudConfig() error {
 	return d.Add(constants.OpCopyCloudConfig,
-		herd.WithDeps(constants.OpPrepareISO),
+		herd.WithDeps(constants.OpPrepareDestination, constants.OpPrepareNetboot, constants.OpPreparetmproot),
 		herd.WithCallback(func(ctx context.Context) error {
 			internal.Log.Logger.Info().Str("cloudConfig", d.Config.CloudConfig).Msg("Copying cloud config")
+			if _, err := os.Stat(d.destination()); err != nil && os.IsNotExist(err) {
```
jimmykarily, line 80: `StepPrepDestination` already does that and it a dependency of this step. Is this what you mean in the description that it has to run at the right point in time? If that's true, what is the purpose of the `StepPrepDestination` ? Maybe what we need is a `herd.Pre` kind of thing that let's us run a function before the actual step runs and we prepare the destination there (maybe it's already supported in herd? I didn't check).

Itxaka: yeah I think I went overboard with extra checks around as it was such a mess. Ideally this should never happen, but that was what we expected before and it still happened. So I added this so we can see the error properly if we mess this up again somehow (we will xD)

PR #315, opened 2025-07-02
Cleanup temp files and directories after ISO building

### LINKED ISSUE #3192: Auroraboot should clean it it stuff
If you build an iso with aurora, the target dir for the iso will get polluted with things from aurora.

config.yaml and build dirs and left in there and they should not be there. Only iso and sha for the iso should be in there

jimmykarily:
Fixes https://github.com/kairos-io/kairos/issues/3192

Doesn't make cleanup optional to keep things simple (no additional config and such). Only cleans up when building an ISO.

Test with:

```bash
rm -rf build/*
docker build -t myauroraboot .
docker run --rm -it -v $PWD/build:/tmp/auroraboot myauroraboot build-iso quay.io/kairos/alpine:3.21-standard-amd64-generic-v3.4.2-k3sv1.32.3-k3s1
ls -liah build
```

there should only be the ISO file and the checksum file.

I'm opening for review. If we want to allow the user to keep the files around with a flag, I'll need to add that.

### DIFF
--- a/pkg/ops/iso.go
+++ b/pkg/ops/iso.go
@@ -758,17 +758,12 @@ func WithImageExtractor(extractor v1types.ImageExtractor) func(r *agentconfig.Co
 }
 
 func cleanupTempFiles(dst string) error {
-	if err := os.RemoveAll(filepath.Join(dst, "temp-rootfs")); err != nil {
-		return fmt.Errorf("cleanining up directory temp-rootfs: %w", err)
+	paths := []string{"temp-rootfs", "netboot", "config.yaml"}
+	for _, path := range paths {
+		fullPath := filepath.Join(dst, path)
+		if err := os.RemoveAll(fullPath); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/ops/iso.go
```diff
@@ -166,6 +166,11 @@ func GenISO(srcFunc, dstFunc valueGetOnCall, i schema.ISO) func(ctx context.Cont
 		if err != nil {
 			internal.Log.Logger.Error().Msgf("Failed generating iso '%s' from '%s'. Error: %s", i.Name, src, err.Error())
 		}
+
+		if err := cleanupTempFiles(dst); err != nil {
```
Itxaka, line 170: there is a step `CleanTmpDirs` that should clean everything, maybe we just have to add it to the deployer and call it a day?

mudler:
Closing in favor of https://github.com/kairos-io/AuroraBoot/pull/316

PR #317, opened 2025-07-07
Bump depos and bring pkcs11 support

Itxaka:
Bump all deps and bring the latest go-ukify which brings pkcs11 support for efi signing

Part of: https://github.com/kairos-io/kairos/issues/3222


mudler (approved):
changes are looking good, however CI is still red

PR #342, opened 2025-08-12
Migrate to alpine js

jimmykarily:
Part of: https://github.com/kairos-io/kairos/issues/3558

Alternative to [htmx + elem-go](https://github.com/kairos-io/AuroraBoot/tree/htmx)

Introduces minimal visual changes.

old form:

image

new form:

image

Same thing for interactions. UX should be almost identical. Only the code has been DRYied and templatized.


jimmykarily:
I made one change in the "base image" field:

image

It's only one value that we need, so it makes sense to be just one input field. Also it's nice because when the user clicks the buttons, it's obvious that there is no magic to it. The user can see immediately which base image they refer to and even edit it to their liking (change the tag or whatever).

internal/web/app/alpine-3.14.8.js
```diff
@@ -0,0 +1,5 @@
..."?o:`${i}.${o}`;typeof s=="object"&&s!==null&&s._x_interceptor?n[o]=s.initialize(e,c,o):t(s)&&s!==n&&!(s instanceof Element)&&r(s,c)})};return r(e)}function Re(e,t=()=>{}){let r={initialValue:void 0,_x_interceptor:!0,initialize(n,i,o){return e(this.initialValue,()=>In(n,i),s=>ht(n,i,s),i,o)}};return t(r),n=>{if(typeof n=="object"&&n!==null&&n._x_interceptor){let i=r.initialize.bind(r);r.initialize=(o,s,a)=>{let c=n.initialize(o,s,a);return r.initialValue=c,i(o,s,a)}}else r.initialValue=n;return r}}function In(e,t){return t.split(".").reduce((r,n)=>r[n],e)}function ht(e,t,r){if(typeof t=="string"&&(t=t.split(".")),t.length===1)e[t[0]]=r;else{if(t.length===0)throw error;return e[t[0]]||(e[t[0]]={}),ht(e[t[0]],t.slice(1),r)}}var ir={};function y(e,t){ir[e]=t}function fe(e,t){let r=Ln(t);return Object.entries(ir).forEach(([n,i])=>{Object.defineProperty(e,`$${n}`,{get(){return i(t,r)},enumerable:!1})}),e}function Ln(e){let[t,r]=_t(e),n={interceptor:Re,...t};return te(e,r),n}function or(e,t,r,...n){try{return r(...n)}catch(i){re(i,e,t)}}function re(e,t,r=void 0){e=Object.assign(e??{message:"No error message given."},{el:t,expression:r}),console.warn(`Alpine Expression Error: ${e.message}
```
mauromorales, line 1: why do we add it hardcoded and not as a package?

internal/web/app/assets/css/tailwind.css
```diff
@@ -1,2 +1 @@
 @import "tailwindcss";
-@import "flowbite/src/themes/default";
```
mauromorales, line 2: why do we remove flowbite?

internal/web/jobstorage/job_storage.go
```diff
@@ -68,6 +68,7 @@ type Artifacts struct {
 type JobData struct {
 	Variant                string    `json:"variant"`
 	Model                  string    `json:"model"`
+	Architecture           string    `json:"architecture"`
```
mauromorales, line 71: important addition hehe 😅

internal/web/app.go
```diff
@@ -125,6 +125,10 @@ func buildHandler(c echo.Context) error {
 	}
 
 	variant := c.FormValue("variant")
+	architecture := c.FormValue("architecture")
+	if architecture == "" {
+		architecture = "amd64" // Default to amd64 if not specified
```
mauromorales, line 130: can we somehow default to host? or by default does it have xcompilation?

internal/web/app/alpine-3.14.8.js
jimmykarily: I was getting random errors, possibly because I was pulling it all the time (restarting the server and such), so I thought it's better if we bundle it. If we are not developing then maybe it's not a big issue. Maybe it was just random cdn hickups. I don't know.

internal/web/app/assets/css/tailwind.css
jimmykarily: because it wasn't used anywhere anymore. It was only for the info popovers which are not implemented with just css and alpinejs, so it's one less depedency (it did remove quite a lot of things from package-lock.json)

internal/web/jobstorage/job_storage.go
jimmykarily: yes :D . It's eventually used as a `--platform` argument on the docker command. I'm not 100% sure this is the was we use it in our pipelines. Is this how auroraboot infers the architecture or is there an explicit argument I should pass? I need to check.

internal/web/app.go
jimmykarily: The architecture of the machine running the build too (auroraboot), doesn't necessarily have anything to do with the machine that is going to run Kairos. Maybe a default value is wrong and we should always require and explicit value from the form. What do you think?

internal/web/app/accordion-view.js
```diff
@@ -0,0 +1,421 @@
...
+            const icons = {
+                'base-image': '📦',
+                'architecture': '🏗️',
+                'model': '🔧',
+                'variant': '⚙️',
+                'kubernetes': '☸️',
+                'kubernetes-release': '📋',
+                'version': '🏷️',
+                'configuration': '⚙️',
+                'artifacts': '📦'
```
mauromorales, line 339: these two are duplicated icons

internal/web/app/assets/css/tailwind.css
mauromorales: noice!

internal/web/app/alpine-3.14.8.js
mauromorales: yeah good question, I don't like the js tooling much haha but I feel if we already have packages is kind of weird that this one we vendor, I'd rather do all the same way and better if it's the package way cuz it's easier to update and such

internal/web/app.go
jimmykarily: done. Made it a mandatory param

internal/web/app/accordion-view.js
```diff
@@ -0,0 +1,421 @@
...
...-4.771 2.83 6.42 3.61 15.621 1.96 18.211 3.801-.62 1.81-2.391 2.989-4.51 3.979 10.34.61 38.609-.38 44.029-22.101C101.796 63.46 82.476 49.61 49.247 32.48zm176.309 0c-33.219 17.13-52.539 30.98-63.119 42.79 5.42 21.721 33.689 22.711 44.029 22.101-2.119-.99-3.889-2.17-4.51-3.979 2.59-1.841 11.791-.19 18.211-3.801-2.471-.51-3.621-1.01-4.771-2.83 6.061-1.93 12.602-3.6 16.441-6.81-2.07.03-4.01.47-6.721-1.41 5.439-2.931 11.23-5.25 15.74-9.729-2.811-.07-5.84-.021-6.721-1.061 4.971-3.08 9.17-6.51 12.641-10.25-3.93.471-5.59.061-6.539-.62 3.76-3.851 8.52-7.1 10.789-11.85-2.92 1.01-5.59 1.39-7.52-.09 1.279-2.88 6.75-4.58 9.9-11.31-3.07.29-6.33.67-6.98 0 1.43-5.82 3.879-9.09 6.279-12.47-6.58-.1-16.549.03-16.1-.53l4.07-4.16c-6.42-1.73-13 .28-17.77 1.77-2.141-1.69.039-3.83 2.648-6.01-5.459.73-10.389 1.98-14.85 3.71-2.379-2.15 1.551-4.3 3.451-6.45-8.441 1.6-12.012 3.85-15.561 6.1-2.58-2.47-.15-4.57 1.59-6.72-6.359 2.36-9.641 5.4-13.09 8.4-1.17-1.58-2.971-2.74-.791-6.54-4.52 2.6-7.92 5.67-10.43 9.1-2.799-1.78-1.67-4.21-1.68-6.45-4.699 3.82-7.68 7.88-11.32 11.85-.73-.54-1.379-2.35-1.949-5.22-11.189 10.86-27.01 38.221-4.061 49.07 19.524-16.11 42.844-27.821 68.694-36.601z" clip-rule="evenodd"/></svg>`,
```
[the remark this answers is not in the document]
jimmykarily: done

internal/web/app/alpine-3.14.8.js
jimmykarily: done. Actually it was already in package.json.

internal/web/app/accordion-view.js
jimmykarily: actually, these icons are legacy code (from previous experiments). They are not used anywhere. Removed along with more unused code.

PR #351, opened 2025-08-26
fix: avoid grub label clash between iso and target

coneheed:
By labeling the grub partition on the ISO as COS_GRUB_LIVE, we avoid any potential installation issues with kairos-agent.

Previously, both the ISO and the target partitions were labeled COS_GRUB and in some circumstances, the wrong partition was selected by the agent

Note - the other hard coded reference to COS_GRUB is in prepare_nvidia_orin_images.sh - but I don't believe this needs to change.

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -43,7 +43,7 @@ const (
 	KairosDefaultArtifactName = "kairos"
 	IsoEFIPath                = "/boot/uefi.img"
 	EfiBootPath               = "/EFI/BOOT"
-	EfiLabel                  = "COS_GRUB"
+	EfiLabel                  = "COS_GRUB_LIVE"
 	EfiFs                     = "vfat"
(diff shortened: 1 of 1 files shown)


pkg/constants/constants.go
```diff
@@ -43,7 +43,7 @@ const (
 	KairosDefaultArtifactName = "kairos"
 	IsoEFIPath                = "/boot/uefi.img"
 	EfiBootPath               = "/EFI/BOOT"
-	EfiLabel                  = "COS_GRUB"
+	EfiLabel                  = "COS_GRUB_LIVE"
```
Itxaka, line 46: This is for a vfat image which seems to only accept max of 8 chars :D

I would change it to maybe EFI_LIVE or something like that, we dont really care that much about its label

Itxaka: I think we can even drop it and dont care about it

PR #352, opened 2025-08-26
fix(nvidia): update fonts path

jasperdekeuk:
Script fails due to not finding `*.pf2` files in `EFI/BOOT`.
Files seem to be in `EFI/BOOT/grub2` instead. 

I guess the path might have changed in a certain version so kept the old path to be backwards compatible.


Itxaka (approved):
lgtm thanks!

mauromorales:
@jasperdekeuk thanks for the contribution!

PR #381, opened 2025-11-08
feat: add support for locating systemd-boot files in source

Itxaka:
- Introduced the `--sdboot-in-source` flag to allow users to specify that systemd-boot files should be searched in the source root filesystem instead of using the bundled files.
- Updated paths for systemd-boot and stub files to reflect the new structure.
- Enhanced error handling for file searches to improve user feedback.

### DIFF
--- a/internal/cmd/build-uki.go
+++ b/internal/cmd/build-uki.go
@@ -1189,7 +1189,7 @@ func FindFirstFileInDir(dir, pattern string) (string, error) {
 			}
 			if matched {
 				foundFile = path
-				return filepath.SkipDir // Stop walking once we found the file
+				return filepath.SkipAll // Stop walking once we found the file
 			}
 
 		}


Itxaka:
This is needed for hadron

internal/cmd/build-uki.go
```diff
@@ -1128,3 +1171,35 @@ func NameFromCmdline(basename, cmdline string) string {
...
+			return err
+		}
+		if !d.IsDir() {
+			matched, err := filepath.Match(pattern, d.Name())
+			if err != nil {
+				return err
+			}
+			if matched {
+				foundFile = path
+				return filepath.SkipDir // Stop walking once we found the file
```
Itxaka:
```suggestion
				return filepath.SkipAll // Stop walking once we found the file
```

PR #493, opened 2026-03-16
feat(sysext): include /opt directory in systemd extensions

Dr-N00B:
By default, auroraboot only includes /usr hierarchy. This change allows
extracting files from /opt in addition to /usr.


Itxaka (commented):
I would do an optional flag as /opt is not supported by default creating sysexts and could lead to big sysexts that ship[ more than the needed stuff

Itxaka (approved):

Itxaka (approved):

PR #497, opened 2026-04-13
Add support for Nvidia Thor cmdline options

Itxaka:
While console options are not strictly needed, the ignore unused options should be always used, otherwise the kernel will shut down those devices and we wont be able to boot properly

### DIFF
--- a/pkg/constants/grub_live_bios.cfg
+++ b/pkg/constants/grub_live_bios.cfg
@@ -9,46 +9,54 @@ if [ -f ${font} ];then
     loadfont ${font}
 fi
 
+# Get model
+smbios --type 4 --get-string 5 --set model
+# For Thor we need to set the ignored clk and pd, otherwise devices will die during boot
+if test $model == "Thor"; then
+  echo "Thor device detected, setting ignored clocks and power domains and proper console output"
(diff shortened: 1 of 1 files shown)


pkg/constants/grub_live_bios.cfg
```diff
@@ -9,46 +9,54 @@ if [ -f ${font} ];then
     loadfont ${font}
 fi
 
+# Get model
+smbios --type 4 --get-string 5 --set model
+# For Thor we need to set the ignored clk and pd, otherwise devices will die during boot
+if test $model == "Thor"; then
+  echo "Thor device detected, setting ignored clocks and power domains and proper console output"
+  set thor_options="pd_ignore_unused clk_ignore_unused console=ttyUTC0,115200 earlycon=tegra_utc,mmio32,0xc5a0000"
```
mauromorales, line 17: does this actually work? nice, I solved this with the template cuz I couldn't make something like this work :( maybe I should move the nomodeset to be similar

https://github.com/kairos-io/AuroraBoot/blob/220faadf8a65d81b3d440358e9dc61ba61ca788d/pkg/ops/iso.go#L335

Itxaka: I think so, just in time options should work easily. The problem with the nomodeset may be that we are not able to know if the arch is arm64 or it may be a different name? we could try to do it on runtime instead

PR #500, opened 2026-04-15
feat: full e2e node build and deployment flow

mudler:
Depends on https://github.com/kairos-io/kairos-agent/pull/1179

To test it e2e, build from the interface an artifact with a Dockerfile among the lines of:

```
FROM golang:1.26-alpine AS agent-builder
RUN apk add --no-cache git
RUN git clone -b feat/daedalus https://github.com/kairos-io/kairos-agent.git /src
WORKDIR /src
RUN CGO_ENABLED=0 go build -o /kairos-agent .

FROM quay.io/kairos/hadron:v0.0.4-core-amd64-generic-v4.0.3
COPY --from=agent-builder /kairos-agent /usr/sbin/kairos-agent

```


jimmykarily:
Something is wrong with the logs box. I see this:

image


but in the logs box things are stuck in:

image

internal/netbootmgr/manager.go
```diff
@@ -0,0 +1,120 @@
...
+	// Verify required files exist.
+	for _, f := range []string{kernel, initrd, squashfs} {
+		if _, err := os.Stat(f); err != nil {
+			return fmt.Errorf("required netboot file not found: %s", f)
+		}
+	}
+
+	// AuroraBoot start-pixie args: <cloud-config> <squashfs> <address> <port> <initrd> <kernel>
+	// Use empty string for cloud-config (not required for netboot).
+	cmd := exec.Command("auroraboot", "start-pixie", "", squashfs, m.address, m.port, initrd, kernel)
```
jimmykarily, line 62: This is the auroraboot repository, does it have to shell out to auroraboot binary?

pkg/auth/middleware.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/kairos-io/AuroraBoot/pkg/store"
+	"github.com/labstack/echo/v4"
+)
+
+// ContextKeyNodeID is the key used to store the authenticated node ID in the echo context.
+const ContextKeyNodeID = "nodeID"
+
+// AdminMiddleware returns an Echo middleware that checks the Authorization header
+// for a Bearer token matching the given admin password.
+func AdminMiddleware(password string) echo.MiddlewareFunc {
```
jimmykarily, line 19: Since we have a database, instead of generating a password and storing it in plain text under `data/secrets/admin-password`, we can have a first time wizard that asks the user what the admin password should be. Then we store it encrypted on database like the common practice is. On login we will only compare the user provided password after encryption to the value stored on the database.

I'm not sure how important it is in this case, since for now it's for local deployments and single user. A user being able to read the plain text file on the disk can as well read the whole sqlite database. But something to consider in the future.

jimmykarily (commented):
a couple of comment to begin with. I'll be making comments as I'm testing things. It's very hard to submit a full review at once.

jimmykarily:
Filtering doesn't seem to work: 
image

jimmykarily:
When you start the netboot server then after a while it shows as not started and lets you start it again. The UX is confusing. Is the server actually running or not? (looking at my vm netbooting, I would say it's still running). If I have multiple artifacts, starting the server on one of them, would stop the previous servers hosting other artifacts?

jimmykarily:
> When you start the netboot server then after a while it shows as not started and lets you start it again. The UX is confusing. Is the server actually running or not? (looking at my vm netbooting, I would say it's still running). If I have multiple artifacts, starting the server on one of them, would stop the previous servers hosting other artifacts?

server logs say:

```
all arguments are required
```

jimmykarily:
At some point, the Node should not be considered "Online", no?

image

jimmykarily:
This is the first round and it will drag forever if we try to fix all the things in this PR. Feel free to merge as is and create separate tickets for my comments or other fixes.

mudler:
> Something is wrong with the logs box. I see this:
> image
> 
> but in the logs box things are stuck in:
> image

how are you running it? this looks like doesn't have enough permissions

mudler:
> This is the first round and it will drag forever if we try to fix all the things in this PR. Feel free to merge as is and create separate tickets for my comments or other fixes.

yes was thinking the same. and I've just added a decommissioning flow when nodes gets deleted and gating via config commands as you raised in the comments. The longer it stays here, more it will grow as I keep noticing UX improvements, better to work from master and start to cut experimental versions out

jimmykarily:
> how are you running it? this looks like doesn't have enough permissions

the error comes from the fact that I was running it without docker (directly with the binary). But the bug I'm referring to is not that. It's the fact that the error message doesn't appear in the log box at all.

jimmykarily:
> > This is the first round and it will drag forever if we try to fix all the things in this PR. Feel free to merge as is and create separate tickets for my comments or other fixes.
> 
> yes was thinking the same. and I've just added a decommissioning flow when nodes gets deleted and gating via config commands as you raised in the comments. The longer it stays here, more it will grow as I keep noticing UX improvements, better to work from master and start to cut experimental versions out

I agree. Feel free to merge and move comments to other PRs

Itxaka (approved):

PR #506, opened 2026-04-26
feat: extraCmdline support

cjntaylor:
Add support for extraCmdline to be used by the web ui so that it can be encorporated in uki builds. This is needed to enable flags like `intel_iommu=on` when running with vfio and was already supported by the cli backend; this just synchronizes the two implementations.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,7 +6,7 @@ services:
       - "${PORT:-9099}:8080"
     volumes:
       - auroraboot-data:/data
-      - /var/run/docker.sock:/var/run/docker.sock
+      - ${AURORABOOT_DOCKER_SOCKET:-/var/run/docker.sock}:/var/run/docker.sock
     environment:
       - AURORABOOT_URL=${AURORABOOT_URL:-}
       - AURORABOOT_ADMIN_PASSWORD=${ADMIN_PASSWORD:-}
--- a/docs/docs.go
+++ b/docs/docs.go
@@ -1150,6 +1150,9 @@ const docTemplate = `{
                 "overlayRootfs": {
                     "type": "string"
                 },
+                "extendCmdline": {
+                    "type": "string"
+                },
                 "provisioning": {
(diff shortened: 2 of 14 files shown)


cjntaylor:
I wasn't sure if you all we're interested in independent PRs, but I needed this for myself so I figured I'd just structure it as a proper PR. Please, do let me know if I've done something untoward or if it needs changes - just want to be helpful. This is a very neat project that I'm gearing up to use in my homelab

docker-compose.yml
```diff
@@ -6,7 +6,7 @@ services:
       - "${PORT:-9099}:8080"
     volumes:
       - auroraboot-data:/data
-      - /var/run/docker.sock:/var/run/docker.sock
+      - /run/user/1000/podman/podman.sock:/var/run/docker.sock
```
jimmykarily: Might not work for everybody. I guess the old code didn't work for you either, but we need a better fix for this.

jimmykarily:
@cjntaylor PRs are more that welcome, thank you for this contribution. I'm moving it to "under review" for someone to have a look.

docker-compose.yml
cjntaylor: Ah, oops, didn't intend to commit that. Good catch, I'll revert

jimmykarily:
The new field value doesn't appear in the "Review" page of the build:

image


Also, I don't think it works. I set a value, generate an ISO, boot it, install it, boot the installed system and the cmdline I added is not there. Something must be missing from the wiring.

mudler-agent:
Closing recommendation after the July 25 refresh request: the feature is still relevant, but this branch is conflicting and the submitted wiring did not pass the maintainer end-to-end test. There has been no author response or branch update, so this PR should be closed as stale and the feature left available for a fresh implementation. A new implementation should cover UI review display, API/store/builder wiring into `uki.Options.ExtendCmdline`, and a round-trip test.

PR #509, opened 2026-05-13
feat: add riscv64 architecture support

mauromorales:
## Summary
- Add riscv64 as a supported architecture for `build-iso`, `sysext`, `unpack`, and `build-uki` commands
- RISC-V boots grub directly via EFI (no shim) - handled in ISO builder
- Add riscv64 container image support using Fedora packages (no luet dependency)
- Add test workflow on `ubuntu-24.04-riscv` to verify native riscv64 builds work
- Add riscv64 to release workflow for multi-arch container images

## Container image for riscv64

The riscv64 stage uses Fedora packages instead of luet:
- `systemd-boot-unsigned` for UKI building
- `grub2-efi-riscv64` for ISO building
- No ARM board firmware (not needed for riscv64)

## Test plan
- [x] CI passes on the riscv64 runner (unit + integration tests)
- [ ] Manual test: `auroraboot build-iso --arch riscv64 <image>` produces a bootable ISO
- [ ] Verify EFI boot path uses `bootriscv64.efi`
- [ ] Test riscv64 container image builds successfully

See kairos-io/kairos#4083 for full riscv64 ecosystem tracking.


jimmykarily (approved):
Do we have hardware to test the artifacts?

mauromorales:
@jimmykarily not that I'm aware of, but my plan is to ping some people to try it out

mauromorales:
I'll have to do the manual testing after we release because cross compilation is not working yet, what works is to take a riscv64 OCI image and create an iso even if host is a different architecture. I've opened https://github.com/kairos-io/kairos/issues/4083 to follow up

PR #512, opened 2026-05-15
fix: add wait-on to Cypress test to prevent race condition

mauromorales:
## Summary

The `test-ui-builder` job was failing intermittently because Cypress started running tests before the web server was ready.

## Changes

- Add `wait-on: 'http://localhost:8080'` to wait for server to be available
- Add `wait-on-timeout: 60` to allow up to 60 seconds for startup
- Remove output redirect so server logs are visible on failure

## Test plan

- [ ] CI `test-ui-builder` job passes

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -263,7 +263,9 @@ jobs:
        uses: cypress-io/github-action@v6
        with:
          working-directory: e2e/web/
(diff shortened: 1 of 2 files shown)


mauromorales:
test-ui-builder is green https://github.com/kairos-io/AuroraBoot/actions/runs/25960722975/job/76315873681?pr=512

PR #519, opened 2026-05-16
fix: configure osv-scanner ignores and run on main

mauromorales:
## Summary

- Add `osv-scanner.toml` to ignore unfixed Moby client CVEs (GO-2026-4883, GO-2026-4887) that have no patched release yet per govulncheck
- Add push trigger to run scanner on main, not just PRs
- Update dependencies

## Context

The Moby/Docker client library has vulnerabilities with "Fixed in: N/A" status. AuroraBoot only uses the Engine API client against operator-controlled hosts, so these are low risk. Ignores expire in 90 days.

## Test plan

- [x] OSV scanner passes with ignores
- [ ] Scanner runs on push to main


mauromorales:
osv scanner is green https://github.com/kairos-io/AuroraBoot/actions/runs/25961175408/job/76316617274?pr=519

PR #520, opened 2026-05-18
chore(builder): change renovate so it can auto-bump the kairos init image

mudler:
Drops the pinned v0.8.4 fallback so fresh builds pick up the most recent kairos-init by default. Explicit pins via opts.KairosInitImage or KAIROS_INIT_IMAGE still win.

### DIFF
--- a/internal/builder/auroraboot/builder.go
+++ b/internal/builder/auroraboot/builder.go
@@ -439,7 +439,7 @@ func (b *Builder) ensureKairosified(ctx context.Context, image string, opts buil
 		kairosInitImage = os.Getenv("KAIROS_INIT_IMAGE")
 	}
 	if kairosInitImage == "" {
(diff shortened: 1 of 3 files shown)


Itxaka (approved):

internal/builder/auroraboot/builder.go
```diff
@@ -439,7 +439,7 @@ func (b *Builder) ensureKairosified(ctx context.Context, image string, opts buil
 		kairosInitImage = os.Getenv("KAIROS_INIT_IMAGE")
 	}
 	if kairosInitImage == "" {
-		kairosInitImage = "quay.io/kairos/kairos-init:v0.8.4"
+		kairosInitImage = "quay.io/kairos/kairos-init:v0.13.0"
```
mudler:
```suggestion
```

PR #534, opened 2026-06-04
fix: fleet-server security hardening (BOLA, TLS, cred-logging, input validation)

wrkode:
## Fleet-server security hardening

Closes the Critical/High blockers from the non-Redfish review (kairos-io/kairos#4117). Independent of the Redfish work; **recommended to merge FIRST** (see note).

### Fixes
| Commit | Fix | Sev |
|---|---|---|
| `1640efc` | **BOLA / node-impersonation** — agent endpoints bound to the authenticated node (route middleware + handler recheck + atomic node-scoped store update); also fixes a route-shadowing bug that made agent REST polling return 401 | Critical |
| `d8e9c03` | **`losetup -D`** global loop-device detach removed (was detaching every host loop device) | Critical |
| `9ce77d2` | **TLS** option (`--tls-cert`/`--tls-key`) + **credential `?token=` redaction** in access logs | High |
| `1a784f3` | kairos-init Dockerfile-RUN **command-injection** input validation | High |
| `d99e5ea` | SecureBoot key-name **path-traversal** validation | High |
| `27fe702` | `UploadOverlay` safe in-process **tar extraction** (containment + symlink reject + size caps) | Med |
| `723663c` | invalid build inputs return **400** (not 500) | Low |

### Validation
- All packages build; unit tests green. The BOLA fix was security-reviewed (ship-it; three-layer enforcement, fails closed, no existence oracle).
- A curl-driven harness exercised every fix against a running server (**24/24 PASS**): cross-node 403s, HTTPS, log redaction, and 400s on malicious build/keyname/overlay input.

### Merge note
This PR and the Redfish PR (kairos-io/kairos#4109) both touch `internal/cmd/web.go` and `pkg/server/server.go` additively. **Merge this one first**; the Redfish PR then rebases onto it (the combined state is already proven in an integration branch — conflicts are small/additive).

### DIFF
--- a/internal/store/gorm/concurrency_test.go
+++ b/internal/store/gorm/concurrency_test.go
@@ -1,0 +1,98 @@
+package gorm_test
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+	"strings"
+	"sync"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	gormstore "github.com/kairos-io/AuroraBoot/internal/store/gorm"
+	"github.com/kairos-io/AuroraBoot/pkg/store"
+)
+
+var _ = Describe("Gorm Store concurrency", func() {
+	var (
+		s   *gormstore.Store
+		ctx context.Context
+	)
+
+	BeforeEach(func() {
+		// Use a file-backed SQLite DB (NOT :memory:) so WAL mode and
+		// busy_timeout are actually exercised — in-memory SQLite has
+		// different locking semantics.
+		dbPath := filepath.Join(GinkgoT().TempDir(), "concurrency.db")
+		var err error
+		s, err = gormstore.New(dbPath)
+		Expect(err).NotTo(HaveOccurred())
+		ctx = context.Background()
+	})
+
+	AfterEach(func() {
+		Expect(s.Close()).To(Succeed())
+	})
+
+	It("does not return 'database is locked' under concurrent command-status writers", func() {
+		const workers = 8
+
+		var (
+			wg       sync.WaitGroup
+			mu       sync.Mutex
+			failures []error
+		)
+
+		for i := 0; i < workers; i++ {
+			wg.Add(1)
+			go func(idx int) {
+				defer GinkgoRecover()
+				defer wg.Done()
+
+				node := &store.ManagedNode{MachineID: fmt.Sprintf("concurrent-%d", idx)}
+				if err := s.Register(ctx, node); err != nil {
+					mu.Lock()
+					failures = append(failures, err)
+					mu.Unlock()
+					return
+				}
+
+				cmd := &store.NodeCommand{ManagedNodeID: node.ID, Command: store.CmdExec}
+				if err := s.CommandCreate(ctx, cmd); err != nil {
+					mu.Lock()
+					failures = append(failures, err)
+					mu.Unlock()
+					return
+				}
+
+				if err := s.UpdateStatusForNode(ctx, cmd.ID, node.ID, store.CommandCompleted, "ok"); err != nil {
+					mu.Lock()
+					failures = append(failures, err)
+					mu.Unlock()
+				}
+			}(i)
+		}
+
+		wg.Wait()
+
+		for _, err := range failures {
+			Expect(strings.ToLower(err.Error())).NotTo(
+				ContainSubstring("database is locked"),
+				"busy_timeout should make writers wait-and-retry instead of failing",
+			)
+		}
+		Expect(failures).To(BeEmpty())
+
+		// All updates must have landed.
+		nodes, err := s.NodeList(ctx)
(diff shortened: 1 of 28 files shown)


Itxaka (approved):

internal/store/gorm/concurrency_test.go
```diff
@@ -0,0 +1,98 @@
...
+	var (
+		s   *gormstore.Store
+		ctx context.Context
+	)
+
+	BeforeEach(func() {
+		// Use a file-backed SQLite DB (NOT :memory:) so WAL mode and
+		// busy_timeout are actually exercised — in-memory SQLite has
+		// different locking semantics.
+		dbPath := filepath.Join(GinkgoT().TempDir(), "concurrency.db")
```
jimmykarily, line 27: needs a cleanup in `AfterEach` ?

pkg/auth/middleware.go
```diff
@@ -51,6 +79,44 @@ func NodeAPIKeyMiddleware(nodeStore store.NodeStore) echo.MiddlewareFunc {
...
+func AgentOrAdminMiddleware(password string, nodeStore store.NodeStore) echo.MiddlewareFunc {
+	return func(next echo.HandlerFunc) echo.HandlerFunc {
+		return func(c echo.Context) error {
+			token := extractBearer(c.Request().Header.Get("Authorization"))
+			if token == "" {
+				return c.JSON(http.StatusUnauthorized, map[string]string{"error": "unauthorized"})
+			}
+			// Admin first — a constant-time-ish equality is fine here; the
+			// admin password is never a valid node API key.
+			if token == password {
```
jimmykarily, line 107: Since we are in a hardening PR, ideally, we wouldn't store the password itself but the password's hash to compare to. The password should only be shown to the user once or selected by the user. The server should never store the password.

We can do this in a separate PR to avoid making this one even bigger.

pkg/handlers/nodes.go
```diff
@@ -348,15 +348,27 @@ func (h *NodeHandler) Heartbeat(c echo.Context) error {
 }
 
 // GetCommands handles GET /api/v1/nodes/:nodeID/commands.
-// For agent requests, returns only pending commands and marks them as delivered.
+//
+// Served under AgentOrAdminMiddleware, so it handles two callers:
```
jimmykarily, line 352: I thought we handled these [as part of a middleware](https://github.com/kairos-io/AuroraBoot/pull/534/changes#diff-65bb2afc855d0e0889576c971861ddfcab632a0de91e320c3bf06b462fc6a1f9R98). Do we need to worry about it in the handlers too?

pkg/server/server.go
```diff
@@ -20,30 +22,79 @@ import (
...
 }
 
 // New creates and configures an Echo server with all routes and middleware.
 func New(cfg Config) *echo.Echo {
 	e := echo.New()
 	e.HideBanner = true
 
-	e.Use(middleware.Logger())
+	// Use a request logger whose URI is run through redactToken before it is
+	// written. All three auth schemes accept the credential as a ?token= query
```
jimmykarily, line 79: Wouldn't it be better if the token was put in a header instead? We would still have to make sure nothing prints it, but headers are generally less likely to appear in logs accidentally.

pkg/server/server.go
```diff
@@ -120,9 +185,10 @@ func New(cfg Config) *echo.Echo {
 	adminGroup.POST("/nodes/:nodeID/decommission", nodeHandler.Decommission)
 	adminGroup.PUT("/nodes/:nodeID/labels", nodeHandler.SetLabels)
 	adminGroup.PUT("/nodes/:nodeID/group", nodeHandler.SetGroup)
-	adminGroup.GET("/nodes/:nodeID/commands", nodeHandler.GetCommands)
+	// GET /nodes/:nodeID/commands and PUT .../commands/:commandID/status are
+	// served by the shared agent-or-admin group above (single registration to
+	// avoid Echo route shadowing); they branch on the caller's identity.
```
jimmykarily, line 190: This comment won't have a point outside this PR. It only explained the remove line. Without the removed line around anymore, the reader won't know what this explains.

pkg/ws/handler_test.go
```diff
@@ -85,12 +86,17 @@ var _ = Describe("WebSocket Handler", func() {
 
 	BeforeEach(func() {
 		var err error
-		// Increment first so every spec gets a fresh, unique DSN — two
-		// specs sharing a name would collide under cache=shared and
-		// cause flaky "database is closed" / "table is locked" races
-		// when the previous spec's handler goroutines haven't drained.
+		// Every spec gets its own private, on-disk SQLite database in a
+		// per-spec temp dir. We deliberately do NOT use a shared-cache
```
jimmykarily, line 90: What do we do to deliberately not use the shared-cache? I don't see anything here.

jimmykarily (commented):
Nice catches! Some small comments here and there.

PR #539, opened 2026-06-08
fix: fleet-server hardening — concurrency (#4117 Medium) + Redfish deploy ISO path

wrkode:
## What

Addresses the four #4117 **Medium** fleet-server concurrency/correctness items deferred from the security-hardening PR (#534), each as an independently reviewable commit, plus a follow-up that closes the last delivery-dedup window.

| Commit | Fix |
|---|---|
| `fix(ws): serialize all writes per WebSocket connection` | `gorilla/websocket` forbids concurrent writes to one `*Conn`, but `Hub.SendCommand` (under shared `RLock`), `UIHub.Broadcast`, the agent read-loop, and the UI ping ticker all wrote to the same conn. New `wsConn` wrapper (conn + per-conn write mutex); every write routes through it. Map locks are no longer held across network writes. |
| `fix(api): deliver each node command exactly once` | Admin-created commands were pushed over WS but left `Pending`, so the next REST poll re-delivered them. New atomic `ClaimForDelivery` (`UPDATE … WHERE phase=Pending`, RowsAffected); push and poll both claim-then-send. |
| `fix(ops): stop mutating process cwd in concurrent build paths` | `os.Chdir` corrupts cwd of concurrent goroutine builds in `web` mode. Eliminated in `InjectISO` (paths already absolute); serialized in `uki.Build` with a documented mutex (third-party go-ukify writes cwd-relative). |
| `fix(server): stop registering handlers on the global DefaultServeMux` | `ServeArtifacts` and the pxe-uki server used `http.Handle("/")` on `DefaultServeMux` → panic on a second registration. Each now uses a private `http.NewServeMux()`. |
| `fix(ws): claim pending commands on connect for exactly-once delivery` | `sendPendingCommands` switched from bulk `MarkDelivered` to per-command `ClaimForDelivery`, closing the last double-delivery window. |
| `test(handlers): return copies from fake GetPending to fix -race` | Test-fake aliasing defect surfaced by the new concurrent-poll test under `-race`; real gorm returns fresh structs per query. |

## Testing

- `go build ./...`, `go vet` (touched packages) — clean.
- `pkg/ws` and `pkg/handlers` green under **`-race`**; `internal/store/gorm`, `pkg/ops`, `pkg/uki`, `pkg/utils` green.
- New tests: 200 concurrent WS writers/conn; store-level + handler concurrent-claim "exactly one winner"; `InjectISO` leaves cwd unchanged; two mux instances without panic.
- Reviewed by an internal staff-engineer pass (production code sound; the one blocker it found — a test-fake `-race` defect — is fixed in the last commit).

## Known follow-ups (intentionally out of scope)

- `uki.Build` serializes concurrent UKI builds end-to-end (the mutex is taken before image extraction) — correctness is fine; throughput could be improved by acquiring it later.
- Claim-before-send is at-most-once on a write failure (command marked `Delivered` but the frame dropped → relies on `ExpiresAt`); consistent across all delivery paths.
- `MarkDelivered` is now unused in production (retained as an interface method).

---

### Added: Redfish deploy ISO-path fix (`145b1a9`)

Bundled here (found while end-to-end testing the fleet server): `DeployRedfish` built the local ISO path as `filepath.Join(artifactsDir, artifactID, isoFile)`, but `ArtifactFiles` already stores each file's **full path** — so the prefix was doubled into a non-existent, relative path and `isoserve.Register` rejected it (`path ... is not absolute`). **Every local-artifact Redfish deploy failed; only explicit `--image-url` worked.** Fixed by resolving with `filepath.Abs`. The handler test modelled `ArtifactFiles` as a bare basename, masking the bug — corrected to the real full-path format, turning the "registers the on-disk ISO" spec into a regression guard. Closes a gap in #4111.

### DIFF
--- a/pkg/handlers/deploy.go
+++ b/pkg/handlers/deploy.go
@@ -291,7 +291,15 @@ func (h *DeployHandler) DeployRedfish(c echo.Context) error {
 				"error": "local ISO serving is not configured on this server; provide imageUrl with a URL the BMC can reach",
 			})
 		}
-		absISO := filepath.Join(h.artifactsDir, artifactID, isoFile)
+		// ArtifactFiles stores each file's path relative to the server's working
+		// directory (e.g. "data/artifacts/<id>/foo.iso"). Resolve it to an
+		// absolute path: isoserve.Register requires an absolute path, and
+		// re-joining it with artifactsDir/artifactID would double the prefix into
+		// a bogus path that does not exist.
+		absISO, err := filepath.Abs(isoFile)
+		if err != nil {
+			return c.JSON(http.StatusInternalServerError, map[string]string{"error": fmt.Sprintf("resolving artifact ISO path: %v", err)})
+		}
 		servedURL, token, err := h.isoServe.Register(absISO, redfishDeployTimeout+5*time.Minute)
 		if err != nil {
 			return c.JSON(http.StatusInternalServerError, map[string]string{"error": fmt.Sprintf("failed to serve artifact ISO: %v", err)})
--- a/pkg/handlers/deploy_test.go
+++ b/pkg/handlers/deploy_test.go
@@ -163,12 +163,16 @@ var _ = Describe("DeployHandler.DeployRedfish", func() {
 		artifactsDir = GinkgoT().TempDir()
 		// Lay down the artifact's on-disk ISO at <dir>/art-1/kairos.iso.
 		Expect(os.MkdirAll(filepath.Join(artifactsDir, "art-1"), 0755)).To(Succeed())
-		Expect(os.WriteFile(filepath.Join(artifactsDir, "art-1", "kairos.iso"), []byte("ISO-BYTES"), 0644)).To(Succeed())
+		isoPath := filepath.Join(artifactsDir, "art-1", "kairos.iso")
+		Expect(os.WriteFile(isoPath, []byte("ISO-BYTES"), 0644)).To(Succeed())
 
 		artifacts = &fakeArtifactStore{}
+		// ArtifactFiles stores each file's FULL path (as the builder writes it),
+		// not a bare basename. Modelling it as a basename previously masked a
+		// path-doubling bug in DeployRedfish, so use the real format here.
 		Expect(artifacts.Create(context.Background(), &store.ArtifactRecord{
 			ID:            "art-1",
-			ArtifactFiles: []string{"kairos.iso"},
+			ArtifactFiles: []string{isoPath},
 		})).To(Succeed())
 		deployments = &fakeDeploymentStore{}
 		bmcTargets = &fakeBMCTargetStore{}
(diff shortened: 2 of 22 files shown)


pkg/handlers/deploy.go
```diff
@@ -291,7 +291,15 @@ func (h *DeployHandler) DeployRedfish(c echo.Context) error {
 				"error": "local ISO serving is not configured on this server; provide imageUrl with a URL the BMC can reach",
 			})
 		}
-		absISO := filepath.Join(h.artifactsDir, artifactID, isoFile)
+		// ArtifactFiles stores each file's path relative to the server's working
+		// directory (e.g. "data/artifacts/<id>/foo.iso"). Resolve it to an
+		// absolute path: isoserve.Register requires an absolute path, and
+		// re-joining it with artifactsDir/artifactID would double the prefix into
```
jimmykarily, line 297: This comment explains why we should not do what the old code did but after the old code is gone the comment won't make much sense anymore.

Also, if the new code is correct, the old code must have been broken. Was it?

pkg/handlers/deploy_test.go
```diff
@@ -163,12 +163,16 @@ var _ = Describe("DeployHandler.DeployRedfish", func() {
...
 		// Lay down the artifact's on-disk ISO at <dir>/art-1/kairos.iso.
 		Expect(os.MkdirAll(filepath.Join(artifactsDir, "art-1"), 0755)).To(Succeed())
-		Expect(os.WriteFile(filepath.Join(artifactsDir, "art-1", "kairos.iso"), []byte("ISO-BYTES"), 0644)).To(Succeed())
+		isoPath := filepath.Join(artifactsDir, "art-1", "kairos.iso")
+		Expect(os.WriteFile(isoPath, []byte("ISO-BYTES"), 0644)).To(Succeed())
 
 		artifacts = &fakeArtifactStore{}
+		// ArtifactFiles stores each file's FULL path (as the builder writes it),
+		// not a bare basename. Modelling it as a basename previously masked a
+		// path-doubling bug in DeployRedfish, so use the real format here.
```
jimmykarily, line 172: This answers my previous comment. But this comment belongs to a commit message not in the code (won't make sense to future readers, unless only in the context of this diff).

pkg/uki/uki.go
```diff
@@ -35,6 +36,16 @@ import (
...
 )
 
+// buildCWDMu serializes the portion of Build that mutates the process working
+// directory. os.Chdir is process-global, and in `web` mode builds run as
+// concurrent goroutines, so two overlapping Build calls would clobber each
+// other's cwd. Eliminating the chdir here is not cleanly possible: the
+// go-ukify Builder writes its UKI to a cwd-relative OutUKIPath and the
+// downstream cpio packing walks the cwd ("."), so both depend on cwd ==
+// sourceDir. Until those are made cwd-independent, we serialize the
+// chdir->Build->restore critical section instead.
```
jimmykarily, line 46: Sounds to me that the proper fix might be simpler but I could be wrong.

jimmykarily (approved):

PR #576, opened 2026-06-23
feat(raw): emit individual partition images for Nvidia flashing

### LINKED ISSUE #3916: feat: Uniform Nvidia image building inside Auroraboot
**Is your feature request related to a problem? Please describe.**
Currently, the build process for creating images for Nvidia is different from other platforms.

**Describe the solution you'd like**
We want to uniform the image building process such as we can use auroraboot directly instead of having hardcoded scripts to prepare the images.

Ideally, Auroraboot should output the partitions as image files, so can later be picked up by flashing tools. For example: efi.img, recovery_partition.img, state_partition.img, oem.img, persistent.img. A raw img file for each partition. See https://kairos.io/docs/installation/nvidia_agx_orin/ for context on the flashing process

**Describe alternatives you've considered**

**Additional context**
#3905

Itxaka:
Add a `disk.partitions` output mode to the EFI raw-disk build. When set, auroraboot emits each partition as its own image file (efi.img, oem.img, recovery_partition.img) into the output dir and skips merging them into a single .raw disk. This lets the Nvidia Jetson AGX Orin flashing workflow use auroraboot directly instead of the standalone
prepare_nvidia_orin_images.sh script, giving a single source of truth for how Kairos partitions are built.

The state/persistent partitions are created on first boot by the existing reset/expand cloud-init baked into the OEM partition, matching the regular raw-disk model.

`disk.partitions` implies an EFI build and is rejected (fail-fast via Config.Validate) when combined with disk.gce/disk.vhd, which convert the merged disk that partitions mode does not produce.

Tests: schema Validate cases, ops emitPartitionImages unit specs, and e2e coverage (partition-image build + validation rejection).

Fixes: kairos-io/kairos#3916


pkg/ops/rawDiskGeneration.go
```diff
@@ -51,6 +51,11 @@ type RawImage struct {
 	elemental            *elemental.Elemental // Elemental instance to use for the operations
 	efi                  bool                 // If the image should be EFI or BIOS
 	config               *sdkConfig.Config    // config to use for the operations
+	// Partitions, when true, emits each partition as a separate image file
+	// (efi.img, oem.img, recovery_partition.img) into Output and skips merging
+	// them into a single .raw disk. Used by flashing workflows such as Nvidia
+	// Jetson AGX Orin. Only valid together with an EFI build.
+	Partitions bool
```
mudler: looks a bit counter-intuitive, maybe better to call it SeparatePartitionsImages so it tells more about what it does ?

mudler (approved):

pkg/ops/rawDiskGeneration.go
Itxaka: sounds good!
@copilot can you rework this to change that Partition into SeparatePartitionsImages please

Itxaka:
@mudler a couple of questions here:
 - will you be able to test this?
 - do we need ALL the partitions? Or can we get away with only this ones and then on boot we do like the rest of the raw images and let the recovery create the state and persistent? Hopefully this is the way of doing it so we can do all the raw partitions the same, it makes the images smaller so less flashing time and we have a siongle way of doing it.

mudler:
> @mudler a couple of questions here:
> 
> * will you be able to test this?

:scream: yes.. sadly I can :D 

> * do we need ALL the partitions? Or can we get away with only this ones and then on boot we do like the rest of the raw images and let the recovery create the state and persistent? Hopefully this is the way of doing it so we can do all the raw partitions the same, it makes the images smaller so less flashing time and we have a siongle way of doing it.

here's the thing, I don't recall perfectly because I've done this long time ago, but IIRC for some reason I could not do partitioning once booted on that emmc controller, as it refused but I don't remember it correctly now, but I went ahead with flashing the expected partitions. I can give it a shot to the new process

Itxaka:
Talked with @mudler here, to not block this and becuase this PR is additive (doesnt break existing workflow for building nvidia) we will merge it now and leave the testing for now as its a bit time consuming, so we cannot guarantee a good timing to test this.

The ticket will still be open until we can confirm as we area also trying a different approach, which means partition happens on the device itself. If that doesnt work, then we will need to create the partitions anyway.

FYI, if this works as android, we may not need to generate the full partition with the stuff, only the partition.xml will need to be there adn we can still reuse the raw stuff of resetting on first boot.

Otherwise we will need to improve this to generate the full state partition which is not bad but boring.

maybe @jasperdekeuk also has interest into this?

jasperdekeuk:
Over the summer we'll have to flash a new batch of Nvidia devices (Orin, Thor, dgx spark), so we can test the new pr's/ workflow

jasperdekeuk:
@Itxaka, @mudler  I have just flashed a Jetson agx orin with the new flow (with `--set "disk.partitions=true"`) and seems to work nicely (after some Nvidia searching pains 😄 ). Some quick notes:

The creation of partitions with `oem/01_reset.yaml` seems to work on the Jetson. 

For the flashing process I used this xml (it is for nvme, not emmc)
So only `efi.img`, `oem.img`, `recovery_partition.img` partition are defined, other partitions are created after first boot reset (as with all other kairos flows). This is a big improvement in speed for flashing! 
```xml
<partition_layout version="01.00.0000">
    <device type="external" instance="0" sector_size="512" num_sectors="EXT_NUM_SECTORS">
        <partition name="master_boot_record" type="protective_master_boot_record">
            <allocation_policy>sequential</allocation_policy>
            <filesystem_type>basic</filesystem_type>
            <size>512</size>
            <file_system_attribute>0</file_system_attribute>
            <allocation_attribute>8</allocation_attribute>
            <percent_reserved>0</percent_reserved>
        </partition>
        <partition name="primary_gpt" type="primary_gpt">
            <allocation_policy>sequential</allocation_policy>
            <filesystem_type>basic</filesystem_type>
            <size>19968</size>
            <file_system_attribute>0</file_system_attribute>
            <allocation_attribute>8</allocation_attribute>
            <percent_reserved>0</percent_reserved>
        </partition>
        <partition name="esp" type="data">
            <allocation_policy>sequential</allocation_policy>
            <filesystem_type>basic</filesystem_type>
            <size>@ESP_SIZE@</size>
            <file_system_attribute>0</file_system_attribute>
            <partition_type_guid>C12A7328-F81F-11D2-BA4B-00A0C93EC93B</partition_type_guid>
            <allocation_attribute>0x8</allocation_attribute>
            <percent_reserved>0</percent_reserved>
            <filename>efi.img</filename>
            <description>**Required.** Contains a redundant copy of CBoot.</description>
        </partition>
        <partition name="COS_OEM" id="20" type="data">
            <allocation_policy>sequential</allocation_policy>
            <filesystem_type>basic</filesystem_type>
            <size>@OEM_SIZE@</size>
            <allocation_attribute>0x8</allocation_attribute>
            <filename>oem.img</filename>
            <description></description>
        </partition>
        <partition name="COS_RECOVERY" id="21" type="data">
            <allocation_policy>sequential</allocation_policy>
            <filesystem_type>basic</filesystem_type>
            <size>@RECOVERY_SIZE@</size>
            <allocation_attribute>0x8</allocation_attribute>
            <filename>recovery_partition.img</filename>
            <description></description>
        </partition>
        <partition name="secondary_gpt" type="secondary_gpt">
            <allocation_policy> sequential </allocation_policy>
            <filesystem_type>basic</filesystem_type>
            <size>0xFFFFFFFFFFFFFFFF</size> <!-- Forces auto-expansion of preceding partition -->
            <file_system_attribute>0</file_system_attribute>
            <allocation_attribute>0x8</allocation_attribute>
            <percent_reserved>0</percent_reserved>
        </partition>
    </device>
</partition_layout>
```


What I do notice though is that the `oem/01_reset.yaml` is creating `ext2` partitions, I do wonder if `ext4` would not make more sense? 

result
```
kairos@kairos-40a7:~$ lsblk -o label,name,size,mountpoint,fstype
LABEL          NAME           SIZE MOUNTPOINT               FSTYPE
COS_ACTIVE     loop0            8G /                        ext2
               mmcblk0       59.2G
               ├─mmcblk0p1   57.8G                          ext4
               ├─mmcblk0p2    128M
               ├─mmcblk0p3    768K
               ├─mmcblk0p4   31.6M
               ├─mmcblk0p5    128M
               ├─mmcblk0p6    768K
               ├─mmcblk0p7   31.6M
               ├─mmcblk0p8     80M
               ├─mmcblk0p9    512K
               ├─mmcblk0p10    64M                          vfat
               ├─mmcblk0p11    80M
               ├─mmcblk0p12   512K
               ├─mmcblk0p13    64M
               ├─mmcblk0p14   400M
               └─mmcblk0p15 479.5M
               mmcblk0boot0     4M
               mmcblk0boot1     4M
               zram0          2.6G [SWAP]
               zram1          2.6G [SWAP]
               zram2          2.6G [SWAP]
               zram3          2.6G [SWAP]
               zram4          2.6G [SWAP]
               zram5          2.6G [SWAP]
               zram6          2.6G [SWAP]
               zram7          2.6G [SWAP]
               zram8          2.6G [SWAP]
               zram9          2.6G [SWAP]
               zram10         2.6G [SWAP]
               zram11         2.6G [SWAP]
               nvme0n1      931.5G
COS_GRUB       ├─nvme0n1p1     64M                          vfat
COS_STATE      ├─nvme0n1p4   47.2G /run/initramfs/cos-state ext2
COS_PERSISTENT ├─nvme0n1p5  868.5G /usr/local               ext2
COS_OEM        ├─nvme0n1p20    64M /oem                     ext2
COS_RECOVERY   └─nvme0n1p21  15.7G                          ext2
```

jasperdekeuk:
Also, `statereset` is now in the grub menu

mudler:
@jasperdekeuk fantastic! thanks for the feedback.

@mudler-agent assemble a PR for the kairos-docs for updating the Nvidia AGX instructions and update https://github.com/kairos-io/kairos/issues/3916

Itxaka:
This is really good news! Also maybe we should generate the xml at the same time? I mean its practically free for us and we know the sizes and whatnot?

> What I do notice though is that the oem/01_reset.yaml is creating ext2 partitions, I do wonder if ext4 would not make more sense?

Yes this is the default. we dont need journaling for example as the system its immutable we gain basically nothing and we tested the movement in the past and found that it caused a bit of problems with expansion and such. It was not worth it fro what we gain, AFAIK all the features that ext4 brougth vs ext2 were useless to us.

Itxaka:
@mudler-agent check if we should generate the xml flashing file at the same time prefilled with the proper izes that we generated

PR #580, opened 2026-06-23
fix(install-agent): POSIX-safe import script and phone-home registration (#4152)

### LINKED ISSUE #4152: AuroraBoot install-agent script fails to register node: heredoc EOF delimiter malformed
### Summary

Registering/signing a Kairos node with AuroraBoot via the `/api/v1/install-agent` curl one-liner fails when the script tries to write `/oem/phonehome.yaml`. The shell reports a malformed here-document and the phonehome config is never written, so the node does not complete registration.

### Steps to reproduce

On a running Kairos node, run the install command from the AuroraBoot UI (or equivalent):

```bash
curl -sfL http://172.39.100.225:9099/api/v1/install-agent \
  | AURORABOOT_URL=http://172.39.100.225:9099 REGISTRATION_TOKEN="devbcn123" sh
```

### Actual behavior

The script starts but fails while writing the phonehome cloud-config:

```
Installing AuroraBoot agent...
Server: http://172.39.100.225:9099
sh: line 59: warning: here-document at line 38 delimited by end-of-file (wanted `EOF')
```

`/oem/phonehome.yaml` is not created (or is incomplete), `kairos-agent start` does not run successfully, and the node never appears in AuroraBoot.

### Expected behavior

The install script should write a valid `/oem/phonehome.yaml`, start `kairos-agent`, and the node should phone home and show up in the AuroraBoot fleet UI.

### Root cause

The bug is in AuroraBoot's `InstallScript` handler (`pkg/handlers/nodes.go`). The here-document that writes `/oem/phonehome.yaml` places the closing `EOF` delimiter on the same line as `${ALLOWED_YAML}`:

```bash
cat > /oem/phonehome.yaml << EOF
#cloud-config
phonehome:
  url: "${AURORABOOT_URL}"
  registration_token: "${REG_TOKEN}"
  group: "${GROUP}"
  allowed_commands:
${ALLOWED_YAML}EOF
```

Shell requires the closing delimiter to be `EOF` alone on its own line. Because the script never emits a line containing only `EOF`, the here-document runs until end-of-file — matching the error above (heredoc opened around line 38, EOF never found by line 59).

The fix should be:

```bash
  allowed_commands:
${ALLOWED_YAML}
EOF
```

### Workarounds

Until AuroraBoot is patched, you can register manually:

```bash
mkdir -p /oem
cat > /oem/phonehome.yaml << EOF
#cloud-config
phonehome:
  url: "http://172.39.100.225:9099"
  registration_token: "devbcn123"
  group: ""
  allowed_commands:
    - upgrade
    - upgrade-recovery
    - reboot
    - unregister
EOF

kairos-agent start
```

Or download the script, fix the heredoc, and run with `bash` instead of piping directly to `sh`.

### Environment

- AuroraBoot fleet server: `http://172.39.100.225:9099`
- Kairos node hostname: `kairos-b843`
- Install method: curl pipe to `sh` (as documented in AuroraBoot UI)
- Related code: [AuroraBoot InstallScript](https://github.com/kairos-io/AuroraBoot/blob/main/pkg/handlers/nodes.go) (`pkg/handlers/nodes.go`, around line 512)

### Notes

Issues are disabled on `kairos-io/AuroraBoot`, so filing here under the AuroraBoot + Kairos phonehome integration track. Happy to open a PR on AuroraBoot if that helps.

mauromorales:
## Summary

- Fix the `/api/v1/install-agent` script so it works when piped to `sh` (POSIX-safe allowed_commands loop, heredoc `EOF` on its own line).
- After `kairos-agent start`, patch the phone-home systemd unit to use the actual `kairos-agent` binary path and verify the service is active — workaround for kairos-agent hardcoding `/usr/sbin/kairos-agent` (kairos-io/kairos#4178).
- Add `sh -n` / `bash -n` syntax checks in unit tests; update integration and Cypress smoke expectations.
- Bump default kairos-init to **v0.14.7** for kairosification builds (includes the upstream agent path fix).

Fixes kairos-io/kairos#4152

## Test plan

- [x] `go test ./pkg/handlers/... ./internal/builder/auroraboot/...`
- [x] Manual: `curl …/install-agent | sh` on Kairos node — phonehome.yaml written, `kairos-agent-phonehome` active, node registers in UI
- [x] `docker compose up --build -d` and rebuild artifact with new kairos-init default

### DIFF
--- a/pkg/handlers/nodes.go
+++ b/pkg/handlers/nodes.go
@@ -503,8 +503,7 @@ IFS=','
 for c in $ALLOWED_RAW; do
     c=$(echo "$c" | xargs)  # trim whitespace
     if [ -n "$c" ]; then
-        ALLOWED_YAML="${ALLOWED_YAML}    - ${c}
-"
+        ALLOWED_YAML="${ALLOWED_YAML}    - ${c}$(printf '\n')"
     fi
 done
 IFS="$_old_ifs"
--- a/pkg/handlers/nodes.go
+++ b/pkg/handlers/nodes.go
@@ -472,9 +472,9 @@ func (h *NodeHandler) GetCommands(c echo.Context) error {
 
 // InstallScript handles GET /api/v1/install-agent.
 func (h *NodeHandler) InstallScript(c echo.Context) error {
-	script := fmt.Sprintf(`#!/bin/bash
+	script := fmt.Sprintf(`#!/bin/sh
 # AuroraBoot agent install script
-# Usage: curl -sSL %s/api/v1/install-agent | AURORABOOT_GROUP=mygroup bash
+# Usage: curl -sSL %s/api/v1/install-agent | AURORABOOT_GROUP=mygroup sh
 
 set -e
 
@@ -498,11 +498,16 @@ GROUP="${AURORABOOT_GROUP:-}"
 # defaults too — deny-all must be configured from the AuroraBoot UI instead.
 ALLOWED_RAW="${AURORABOOT_ALLOWED_COMMANDS:-upgrade,upgrade-recovery,reboot,unregister}"
 ALLOWED_YAML=""
-IFS=',' read -ra _cmds <<< "$ALLOWED_RAW"
-for c in "${_cmds[@]}"; do
+_old_ifs="$IFS"
+IFS=','
+for c in $ALLOWED_RAW; do
     c=$(echo "$c" | xargs)  # trim whitespace
-    [ -n "$c" ] && ALLOWED_YAML="${ALLOWED_YAML}    - ${c}"$'\n'
+    if [ -n "$c" ]; then
+        ALLOWED_YAML="${ALLOWED_YAML}    - ${c}
+"
+    fi
 done
+IFS="$_old_ifs"
 
 echo "Installing AuroraBoot agent..."
 echo "Server: ${AURORABOOT_URL}"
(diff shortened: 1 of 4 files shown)


pkg/handlers/nodes.go
```diff
@@ -528,7 +533,23 @@ if ! command -v kairos-agent >/dev/null 2>&1; then
 fi
 echo "Starting kairos-agent..."
 kairos-agent start
-echo "kairos-agent-phonehome service installed and started."
+
+# kairos-agent writes the phone-home unit with a hardcoded /usr/sbin/kairos-agent
+# path, but some images install the binary elsewhere (e.g. /usr/bin). systemd
+# then fails with status=203/EXEC and the node never registers.
+AGENT_BIN="$(command -v kairos-agent)"
```
mudler, line 540: while I'm fine to this guard, I'm wondering what actually broke there? are we installing it in different places?

mudler (approved):

pkg/handlers/nodes.go
Itxaka: https://github.com/kairos-io/kairos/issues/4152

Itxaka (approved):

Itxaka (approved):

pkg/handlers/nodes.go
mudler: that doesn't tell about `some images install the binary elsewhere (e.g. /usr/bin)`

mauromorales: Yeah good observation, I guess the alternative would be to hardcode `/usr/bin` cuz it seems that's where we install it always:

https://github.com/kairos-io/kairos-init/blob/main/pkg/stages/steps_install.go#L462

PR #602, opened 2026-06-30
Maas curtin hooks

jimmykarily:
Part of: https://github.com/kairos-io/kairos/issues/3808

This is the part that builds a raw image suitable for MaaS. MaaS looks for a partition that has curtin hooks and chroots to that. I tried to re-use COS_OEM for that but there are checks for certain tools in that partition (e.g. cloud-init). If we added those in COS_OEM, they would survice the first ephemeral boot and would pollute COS_OEM with garbage. I opted for a new partition to keep things tidy.

The produced image can be uploaded to a MaaS server.

Related PRs:

- https://github.com/kairos-io/kairos-init/pull/388
- https://github.com/mudler/yip/pull/302


jimmykarily:
Replace by: https://github.com/kairos-io/AuroraBoot/pull/603 (to let CI run)

PR #615, opened 2026-07-03
4055 webui operator builds

jimmykarily:
Implements https://github.com/kairos-io/kairos/issues/4055

Needs: https://github.com/kairos-io/kairos-operator/pull/142


pkg/handlers/artifacts.go
```diff
@@ -424,6 +484,127 @@ func (h *ArtifactHandler) GetLogs(c echo.Context) error {
...
+//	@Failure	401	{object}	APIError
+//	@Failure	404	{object}	APIError
+//	@Failure	413	{object}	APIError
+//	@Router		/api/v1/artifacts/{id}/upload/{filename} [put]
+func (h *ArtifactHandler) Upload(c echo.Context) error {
+	id := c.Param("id")
+	filename := c.Param("*")
+	ctx := c.Request().Context()
+
+	if err := safePathSegment(id); err != nil {
```
wrkode, line 526: **[Security · High] `Upload` is not phase-gated and the per-build token never expires.**

`Upload` accepts a valid-token `PUT` regardless of the record's phase — including after the build is `Ready`. A token holder can therefore overwrite `kairos.iso` (and its `.sha256`) on an already-finished build at any later time; since nodes/BMCs consume these artifacts, that's an artifact/supply-chain-poisoning vector. The token is also readable by anyone with `get secrets` in the builder namespace (`<id>-upload` Secret).

Suggest: reject uploads once the build is in a terminal phase (`Ready`/`Error`), and invalidate the token (clear `UploadToken`) when `watchCRPhase` observes a terminal phase. Optionally add a `UploadTokenExpiresAt`.

pkg/handlers/artifacts.go
```diff
@@ -424,6 +484,127 @@ func (h *ArtifactHandler) GetLogs(c echo.Context) error {
...
+
+	// Reflect the newly-uploaded file in the store record so the UI's
+	// existing artifact list (which reads ArtifactFiles verbatim) surfaces
+	// operator-produced artifacts the same way it does local ones. Dedupe
+	// the append so a retried exporter Job (backoffLimit) does not stack
+	// duplicates. Ignore Update failures - the file is on disk, the store
+	// row can be reconciled later, and returning 500 here would mislead the
+	// exporter into retrying an already-durable upload.
+	appendArtifactFile(rec, clean)
+	_ = h.store.Update(ctx, rec)
```
wrkode: **[Correctness · Medium] This full-row `Update` re-introduces the exact clobber the store fix removed.**

`ArtifactUpdate` is `Omit("logs").Save(rec)` — a full-row save. `rec` was snapshotted at handler entry, so this writes back a **stale `phase`/`message`**, racing `watchCRPhase`'s column-scoped `ArtifactUpdatePhaseMessage`. Because `watchCRPhase` only re-writes on a *phase change*, the reverted value can persist until the next genuine transition. It self-heals at the terminal write only because the operator marks `Ready` after the last upload — an unstated ordering invariant with no test, and it contradicts the interface doc that says state publishers must prefer the column-scoped update over `Update`.

Suggest: add a column-scoped `ArtifactFiles` update (mirror `ArtifactUpdatePhaseMessage`) and use it here instead of `Save`; add a test interleaving `Upload` with a phase write.

internal/builder/operator/exporter.go
```diff
@@ -0,0 +1,80 @@
...
+					{
+						Name:    "upload-to-auroraboot",
+						Image:   curlUploaderImage,
+						Command: []string{"sh", "-ec"},
+						Args: []string{`
+for f in /artifacts/*; do
+  [ -f "$f" ] || continue
+  base=$(basename "$f")
+  echo "Uploading $base to $AURORABOOT_URL"
+  curl -fsSL --retry 3 -X PUT \
```
wrkode: **[Security · Medium] Token + artifact shipped over plaintext, and `-L` is unnecessary.**

The exporter PUTs the artifact with `Authorization: Bearer <token>` to `AuroraBootURL`, which defaults to `http://<hostname>` with no scheme enforcement. On a plaintext deployment the token and the entire artifact traverse the cluster network in the clear (token capture → feeds the overwrite vector above; artifact tampering).

Suggest: require `https://` (or loudly warn + refuse to inject the exporter) when `--builder=operator` over plaintext. Separately, drop `-L` from `curl -fsSL` — the upload target is fixed, so following redirects only adds a (small) credentialed-redirect risk.

pkg/store/store.go
```diff
@@ -177,6 +177,13 @@ type ArtifactRecord struct {
 	OverlayRootfs           string    `json:"overlayRootfs,omitempty"`
 	ArtifactFiles           []string  `json:"artifacts" gorm:"serializer:json"`
 	Logs                    string    `json:"-" gorm:"type:text"`
+	// UploadToken is a per-build bearer used by the operator backend's
+	// exporter Job to push artifacts to PUT /api/v1/artifacts/:id/upload/:file.
+	// Minted on Create for every build regardless of backend and never
+	// serialized to clients: the JSON tag stays "-" so a compromised UI
+	// bearer cannot lift it, and only the operator-materialized Secret ever
+	// carries it in cleartext.
+	UploadToken string `json:"-"`
```
wrkode, line 212: **[Security · Medium] Upload token is persisted in plaintext, unlike BMC creds.**

`json:"-"` correctly keeps it out of API responses (good), but it's stored plaintext in the DB while BMC credentials go through the cipher. DB read access (backup, the SQLite file, SQLi elsewhere) then yields live write tokens for every build.

Suggest: store a hash of the token and compare hashes in `Upload` (cleanest — the plaintext only needs to live in the k8s Secret), or encrypt with the existing cipher.

internal/builder/operator/logstream.go
```diff
@@ -0,0 +1,403 @@
...
+// the buildah/auroraboot pipeline produced them; broadcasts carry the same
+// prefixed text so subscribers render matching output.
+type broadcastingSink struct {
+	ctx         context.Context
+	buildID     string
+	store       store.ArtifactStore
+	broadcaster builder.LogBroadcaster
+}
+
+func (b *broadcastingSink) WriteLine(container, line string) error {
```
wrkode, line 357: **[Security · Medium] Streamed pod logs can leak the injected cloud-config secret.**

This forwards every container of every build-labeled pod into the persisted store (`AppendLog`) and the UI WebSocket. The injected cloud-config Secret contains the fleet `registration_token`, group, and the default node password. If buildah/kairos-init or any build step echoes the mounted cloud-config (common in debug output), those secrets land in `GET /api/v1/artifacts/:id/logs` and the live log pane in plaintext.

Suggest: scrub known secret values (registration token, cloud-config body) from log lines before persist/broadcast — don't rely on the operator never echoing them. At minimum, document the risk.

internal/builder/operator/logstream.go
```diff
@@ -0,0 +1,403 @@
...
+	"github.com/kairos-io/AuroraBoot/pkg/builder"
+	"github.com/kairos-io/AuroraBoot/pkg/store"
+)
+
+const (
+	// podDiscoveryBudget is how long we wait for the operator to create the
+	// build Pod before giving up. Five minutes is generous enough to cover a
+	// slow operator reconcile loop but not so long that an RBAC/CRD failure
+	// keeps the streaming goroutine alive forever.
+	podDiscoveryBudget       = 5 * time.Minute
```
wrkode: **[Correctness · Low] `waitForPod` / `podDiscoveryBudget` are dead on the production path.**

`streamPodLogs` → `streamAllArtifactPods` never calls `waitForPod` and never applies `podDiscoveryBudget`; it loops until ctx cancellation. So the documented "give up after 5 min so an RBAC/CRD failure can't keep the streaming goroutine alive forever" guarantee no longer holds — termination now depends entirely on `watchCRPhase` seeing a terminal phase or NotFound. `waitForPod` is exercised only by tests.

Suggest: either delete the now-dead func/const, or bound `streamAllArtifactPods` with a discovery budget so a wedged `Building`-forever CR with no Cancel can't leak a 2s-polling goroutine.

internal/builder/operator/translate.go
```diff
@@ -0,0 +1,176 @@
...
+	kairosVersion := firstNonEmpty(opts.Source.KairosVersion, opts.KairosVersion)
+
+	// Pre-built means: consume the caller's ref as-is, no kairos-init pass.
+	// Any signal that we should do build-time work (a Dockerfile, an explicit
+	// KairosVersion, or a KairosInitImage override) drops us out of the
+	// pre-built branch so the caller's intent lands on the operator's
+	// kairos-init flow rather than getting silently ignored. This mirrors
+	// the local backend's ensureKairosified behaviour, which always
+	// kairosifies unless the image is already Kairos-tagged.
+	preBuilt := baseRef != "" &&
```
wrkode, line 49: **[Correctness · Low] `preBuilt` silently drops `KubernetesDistro`/`KubernetesVersion`.**

A request of "prebuilt base + KubernetesDistro/Version" takes the `preBuilt` branch and consumes the ref as-is — the k8s knobs never reach `BuildOptions`, whereas the local backend's `ensureKairosified` would run kairos-init with `-p <distro>`. The `Build` doc comment lists silently-ignored fields but omits these.

Suggest: confirm this is intended (prebuilt-means-prebuilt) and document it, or fold `KubernetesDistro` into the build-time-intent signals.

internal/builder/operator/builder.go
```diff
@@ -0,0 +1,387 @@
...
+
+	// Detach from the request context: the HTTP handler's ctx is cancelled
+	// once the response returns, but the build (and its log stream) outlives
+	// that. Cancel is wired via b.active so the streaming goroutine still
+	// stops promptly when the user cancels the build.
+	streamCtx, cancel := context.WithCancel(context.Background())
+	b.activeMu.Lock()
+	b.active[id] = cancel
+	b.activeMu.Unlock()
+	go b.streamPodLogs(streamCtx, id)
```
wrkode: **[Correctness · Low] `Build` goroutines leak in unit tests.**

Every `Build` starts `streamPodLogs` + `watchCRPhase` on a `context.Background()`-derived ctx. The store-less `Build` unit tests never `Cancel`, so `streamPodLogs` keeps polling the fake clientset every 2s until the process exits — a goroutine leak and latent flake source (harmless to prod correctness).

Suggest: `DeferCleanup(cancel)` / goleak in the tests; the L1 discovery budget would also bound this.

wrkode (commented):
## Review — operator-backed ArtifactBuilder

Thanks for this — it's a careful, well-tested feature and the abstraction is clean (sentinel errors → 400/501, `LogBroadcaster` hoisted to break the import cycle, compile-time interface assertions on both backends). `go build`/`vet`/unit tests are green and the e2e suite is properly gated. I ran it past a security threat-model pass and a correctness pass; findings inline.

**Both passes converged on the `Upload` handler**, which is the one spot I'd ask to address before this hardens:
- it accepts a valid-token PUT regardless of build phase (artifact-overwrite / supply-chain vector), and
- its full-row `Save` re-introduces the phase/message clobber that this PR's own store fix (`41ab3e2`) removed.

A single change fixes both: a column-scoped `ArtifactFiles` update + a terminal-phase gate + token invalidation on terminal.

**Also worth a look (Medium):** the exporter ships the token+artifact over plaintext `http://` by default; the token is stored plaintext at rest (vs. encrypted BMC creds); and streamed pod logs can surface the injected cloud-config secret (registration token, node password) in `/logs` and the UI.

**Confirmed solid:** path traversal is fully contained (`safePathSegment` on id *and* filename, both routes); constant-time token compare; `system/builder` is admin-authed with `sanitizeClusterURL` stripping creds; Secrets owner-ref GC with the CR; overlay tar rejects symlinks/abs/parent with size caps; the store lost-update fix is correct; tests are behavioral (real label-selector filtering, no `time.Sleep` sync).

Everything below the two Mediums is Low/nit and fine as follow-ups. Nice work overall.

Itxaka:
Reviewed at a high level — big PR, mostly looks solid on the security and concurrency side (path traversal guards, constant-time token compare, `defer wg.Wait()` in the log streamer, kubeconfig sanitization, orphan CR cleanup on secret-create failure). Two things stood out that I think are worth addressing before merge:

1. **`internal/builder/operator/exporter.go:52` — curl upload has no timeout.** If the AuroraBoot server becomes unreachable mid-upload, the exporter Job pod hangs until the cluster evicts it. A `--max-time` (and maybe `--connect-timeout`) on the curl call would bound the failure and let the Job fail cleanly so the operator can retry / report.

2. **`pkg/store/store.go:180` — `UploadToken` column added to `ArtifactRecord` without an explicit migration path.** Fine if GORM `AutoMigrate` is guaranteed to run on startup for every deployment, but worth double-checking — existing databases will otherwise hit "column not found" on the first upload with a confusing error. If AutoMigrate covers it, no action needed; if not, backfill needs a migration step.

Everything else I looked at (logstream reconnect, broadcaster, secret handling, /system auth, artifact upload token flow, path segment validation) looked correct to me. Nice work on a chunky PR.

Itxaka (approved):

internal/builder/operator/logstream.go
```diff
@@ -0,0 +1,403 @@
...
+			go func() {
+				defer wg.Done()
+				if err := streamAll(ctx, src, &pod, sink, retryInterval, maxRetries); err != nil {
+					if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+						return
+					}
+					_ = sink.WriteLine("auroraboot", fmt.Sprintf("warning: log streaming for pod %q ended: %v", pod.Name, err))
+				}
+			}()
+		}
```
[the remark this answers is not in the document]
jimmykarily: Fixed: pass `pod` by value to the goroutine. `62d9228`

pkg/handlers/artifacts.go
jimmykarily: Fixed: reject upload when phase is terminal, and `watchCRPhase` clears the upload token on Ready/Error. `869ff14`

pkg/handlers/artifacts.go
jimmykarily: Fixed: column-scoped `UpdateFiles` replaces the full-row Save; regression test interleaves the file write with a phase transition. `869ff14`

internal/builder/operator/exporter.go
jimmykarily: Fixed: dropped `-L` from the curl call and added a startup warning when `--builder=operator` is paired with an http `--url`. `feb47e6`

pkg/store/store.go
jimmykarily: Fixed: DB stores `sha256(token)`; plaintext lives only in the k8s Secret. Upload hashes the presented bearer before the constant-time compare. `0325afa`

internal/builder/operator/logstream.go
jimmykarily: Fixed: added `RedactLine` and populate `LogRedactValues` with the fleet registration token and the node password; both the operator sink and the local dbLogWriter scrub every chunk before persist/broadcast. Best-effort (strict substring, ≥8 chars). `6cdad1d`

internal/builder/operator/logstream.go
jimmykarily: Fixed: deleted `waitForPod`, `podDiscoveryBudget`, and their tests. Doc comment on the surviving poll interval calls out that termination is `watchCRPhase`s job. `62d9228`

internal/builder/operator/builder.go
jimmykarily: Fixed: added a `cancelOnCleanup` helper and call it after each successful `Build`. `62d9228`

internal/builder/operator/translate.go
jimmykarily: Deliberately not folded in. kairos-init installs its provider binaries only in a Variant=Standard from-scratch build (`main.go:40-57`, `steps_install.go:534` in kairos-init) and the local backend's `ensureKairosified` short-circuits on `/etc/kairos-release`, so "prebuilt Kairos + KubernetesDistro=k3s" silently drops the k8s knob on **both** backends today. Rejecting the combination outright is the right long-term fix but belongs to a shared change across both backends. Documented as a known drop in `translate.go`. `62d9228`

jimmykarily:
Thanks! Both items addressed:

- **Curl timeout:** added `--connect-timeout 30` and `--max-time 3600` on the exporter's curl. `0570e45`
- **UploadToken migration:** already covered — `gorm.New` calls `AutoMigrate(&ArtifactRecord{}, ...)` on every startup, and GORM's AutoMigrate adds missing columns via `ALTER TABLE`. Existing databases pick up the new column on next boot with no manual step.

mudler:
@mudler-agent review

jimmykarily:
If @mudler-agent comes up with comments, I'll address as follow-ups. Merged.

PR #646, opened 2026-07-22
fix(builder): derive prebuilt base images per artifact

### LINKED ISSUE #4148: AuroraBoot: standard Hadron build reuses core quay tag; pre-built bases should still produce a per-artifact image
### Summary

When building ISOs with AuroraBoot using a pre-built Hadron image from `quay.io/kairos/hadron`, setting **variant: standard** (with Kubernetes options) does not change the container image used for the build. The artifact still uses the same `-core-` quay tag as a core build. That is a bug: variant and build settings are ignored for pre-built Kairos images.

### How I found it

I built two Hadron artifacts, intending one as core and one as standard:

| Artifact | Name | UI variant | `container_image` | ISO filename |
|----------|------|------------|-------------------|--------------|
| `1ca7563a-...` | DevBcn v1 | core | `quay.io/kairos/hadron:v0.2.0-core-amd64-generic-v4.1.0` | `kairos-hadron-...-**core**-...iso` |
| `4f27d377-...` | DevBcn v2 | **standard** | **same** `quay.io/kairos/hadron:v0.2.0-core-amd64-generic-v4.1.0` | **same** `...-**core**-...iso` |

The review screen for the standard build showed **Variant: standard** and K8s settings, but **Image source** still pointed at the core quay tag. Both builds ended up with the same `container_image` and the same core ISO.

Build logs for the standard artifact still showed:

```
=== Using base image: quay.io/kairos/hadron:v0.2.0-core-amd64-generic-v4.1.0 ===
Image: quay.io/kairos/hadron:v0.2.0-core-amd64-generic-v4.1.0
```

### Contrast with Ubuntu (same AuroraBoot setup)

With `ubuntu:24.04`, each build produces a dedicated local image and variant changes the result:

- Input stays `ubuntu:24.04`
- Built image: `auroraboot-kairos:<artifact-id>` (visible in `docker images`)
- Core vs standard differ in kairos-init (e.g. `-p k3s` for standard) and in the ISO output

Examples from my environment:

- `651cf1ca-...` (core): `container_image: auroraboot-kairos:651cf1ca-...`
- `da87ab16-...` (standard): `container_image: auroraboot-kairos:da87ab16-...`

Hadron on quay does not follow this pattern today.

### Current behavior (bug)

For pre-built images that already contain `/etc/kairos-release`, AuroraBoot skips kairos-init and uses the pulled quay tag directly as `container_image`. UI fields such as **variant** and **Kubernetes** are stored on the artifact record but do not change which image is built or referenced.

So a standard build can silently produce the same core image and ISO as a core build when the same quay tag is used.

### Expected behavior

Even when the base is a pre-built image pulled from quay, AuroraBoot should **always** produce a per-artifact build (e.g. `auroraboot-kairos:<artifact-id>`), the same way it does for plain bases like `ubuntu:24.04`.

The quay image is only the starting point. The final ISO includes cloud-config and other build-time inputs that are not in the pre-built container. The artifact's container image should reflect that derived result—not a direct pointer to the unchanged quay tag.

Concretely:

1. **Variant standard must not reuse a core tag** — standard/K8s settings must affect the built image, not only metadata.
2. **Every artifact build should get its own `auroraboot-kairos:<artifact-id>` image**, whether the base is `ubuntu:24.04` or a pre-built `quay.io/kairos/hadron:*` tag.

### Environment

- AuroraBoot via docker-compose, Docker socket mounted
- Base used: `quay.io/kairos/hadron:v0.2.0-core-amd64-generic-v4.1.0`
- Artifact IDs: `1ca7563a-0a0b-48a9-8612-f927ada42997` (core), `4f27d377-3bdb-4b95-9b0b-ec98afe41f1c` (standard, same container tag)

mudler-agent:
Fixes kairos-io/kairos#4148.

## What

Direct base-image builds now always pass through the existing kairos-init derivation and produce `auroraboot-kairos:<artifact-id>`. Dockerfile builds that already produced a complete Kairos image keep their existing shortcut.

The generated build ID is also copied into `BuildOptions` before the asynchronous build starts, so callers that omit an ID still receive a valid derived image tag.

## Why

`ensureKairosified` returned any image containing `/etc/kairos-release` unchanged. A pre-built Hadron core tag therefore bypassed the variant and Kubernetes flags, so a requested standard artifact reused the original core tag and produced a core ISO.

## How

The Kairos-image shortcut is now limited to Dockerfile outputs, which are already per-artifact images. Direct bases use the existing generated Dockerfile and docker build path, including provider flags and the per-artifact tag.

Regression tests use a fake docker executable to verify:

- a pre-built Hadron base is derived and receives k3s flags
- generated artifact IDs reach the derived image tag
- complete Dockerfile-built Kairos images are not built a second time

## Verification

- `go test ./internal/builder/auroraboot -count=1`
- CI-equivalent unit/integration suite: `go run github.com/onsi/ginkgo/v2/ginkgo -r -p --covermode=atomic --coverprofile=coverage.out --timeout=2h --skip-package ./e2e ./...`
- `git diff --check`

### DIFF
--- a/internal/builder/auroraboot/builder.go
+++ b/internal/builder/auroraboot/builder.go
@@ -297,8 +297,15 @@ func (b *Builder) run(ctx context.Context, bs *buildState, opts builder.BuildOpt
 		logWriter.Flush()
 	}
 
-	// Step 1.5: Ensure image is kairosified.
-	containerImage, err := b.ensureKairosified(ctx, containerImage, opts, outputDir, logWriter)
+	// Step 1.5: Direct base images must always be derived so requested provider
+	// and variant settings are applied. Dockerfile images only need derivation
+	// when the Dockerfile did not produce a complete Kairos image.
+	var err error
+	if opts.Dockerfile == "" {
+		containerImage, err = b.kairosify(ctx, containerImage, opts, outputDir, logWriter)
+	} else {
+		containerImage, err = b.ensureKairosified(ctx, containerImage, opts, outputDir, logWriter)
+	}
 	if err != nil {
 		msg := fmt.Sprintf("kairosify failed: %v", err)
 		b.setPhase(bs, builder.BuildError, msg)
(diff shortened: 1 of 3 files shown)


mudler (approved):

internal/builder/auroraboot/builder.go
```diff
@@ -497,10 +498,13 @@ func (b *Builder) ensureKairosified(ctx context.Context, image string, opts buil
 	if err := validateKairosInitOptions(opts); err != nil {
 		return "", fmt.Errorf("validating kairos-init options: %w", err)
 	}
-	if b.isKairosified(ctx, image) {
+	// dockerBuild already creates a per-artifact image. Preserve the existing
+	// shortcut for Dockerfiles that produce a complete Kairos image, but never
+	// return a direct base-image tag because its requested variant and provider
+	// settings have not been applied yet.
+	if opts.Dockerfile != "" && b.isKairosified(ctx, image) {
```
mudler: but this seems to put the business logic in the wrong place - why we check if there has to be a dockerfile here? better to be defensive here, and instead we should explicitly build the image in the case of the core images, because these are an exception

mudler (approved):

Itxaka (approved):

PR #657, opened 2026-07-24
fix: propagate hostname updates from agents to the UI

### LINKED ISSUE #4196: AuroraBoot: node hostname stuck at registration-time default (e.g. kairos) instead of updating
## Summary

In the AuroraBoot UI, a node's **hostname is captured once at registration and never updated**, even when the machine's actual hostname changes shortly after first boot.

This is especially visible when cloud-config sets a templated hostname (e.g. `kairos-{{ trunc 4 .MachineID }}` → `kairos-a1b2`) but phone-home registration happens **before** that hostname is applied. The node then appears in the UI as the generic default (`kairos`) for its entire lifetime.

## Steps to reproduce

1. Build/boot a Kairos image with AuroraBoot phone-home enabled and a cloud-config hostname template (e.g. `hostname: kairos-{{ trunc 4 .MachineID }}`).
2. Let the node register with AuroraBoot on first boot.
3. Once cloud-init has applied the final hostname, check the AuroraBoot **Nodes** UI.

## Actual behavior

- The UI shows the hostname from the initial registration (often `kairos`).
- The stored value does not change on subsequent heartbeats or reconnects.

## Expected behavior

The AuroraBoot UI should reflect the **current hostname reported by the node**, updating when the machine's hostname changes after registration.

## Root cause (current implementation)

Hostname is a one-shot field today:

1. **kairos-agent** reads `os.Hostname()` only when calling `Register` and sends it in the registration payload.
2. **AuroraBoot** stores `req.Hostname` only when creating a new node record. On idempotent re-register (same `machineID`), it returns the existing credentials and **does not update hostname**.
3. **Heartbeats** (REST and WebSocket) update `agentVersion` and `osRelease` but **do not carry or persist hostname**.

Relevant code:

- Agent registration: [kairos-agent `internal/phonehome/client.go`](https://github.com/kairos-io/kairos-agent/blob/main/internal/phonehome/client.go) — `Register()` uses `os.Hostname()` once
- AuroraBoot registration: [AuroraBoot `pkg/handlers/nodes.go`](https://github.com/kairos-io/AuroraBoot/blob/main/pkg/handlers/nodes.go) — `Register()` sets `Hostname: req.Hostname` only on first insert; re-register path returns early without update
- AuroraBoot heartbeat: [AuroraBoot `pkg/handlers/nodes.go`](https://github.com/kairos-io/AuroraBoot/blob/main/pkg/handlers/nodes.go) — `Heartbeat()` / WS heartbeat update version/OS release only

## Proposed fix

- Include **current hostname** in periodic agent reports (heartbeat and/or re-register).
- AuroraBoot should **upsert hostname when it changes** (compare with stored value; update record + broadcast to UI via existing polling/WS if applicable).
- Optionally delay first registration until after hostname is stable — but syncing on heartbeat is simpler and also covers legitimate hostname changes later in the node's life.

## Acceptance criteria

- [ ] Node registered while hostname is still `kairos` updates to `kairos-xxxx` (or whatever cloud-config applied) once the agent observes the change.
- [ ] UI node list and node detail page show the updated hostname without manual intervention.
- [ ] Re-register / heartbeat path updates hostname idempotently (no duplicate nodes, no spurious churn when hostname is unchanged).

## Related

- kairos-io/kairos#4195 — batch upgrade resilience (same AuroraBoot fleet UX area)

Itxaka:
Nodes showed the boot-time default hostname (`kairos`) in the UI forever because the server only stored `Hostname` on first register. Re-registers with the same `machineID` returned early with the existing credentials, and REST/WS heartbeats carried `agentVersion` + `osRelease` but no hostname. Cloud-config templated names like `hostname: kairos-{{ trunc 4 .MachineID }}` applied by cloud-init after the first phone-home never reached the UI.

Server side:

- Add `hostname` to the REST and WS heartbeat payloads and to `APIHeartbeatRequest`. Empty preserves, non-empty upserts — same pattern as `addresses`/`bootState`, so an older agent that omits the field never wipes a stored value.
- Extend `NodeStore.UpdateHeartbeat` with the new field; gorm store upserts only when non-empty.
- Add `NodeStore.SetHostname` for a lightweight column update with no heartbeat/phase side effects, so the REST Register handler can refresh the hostname on an idempotent re-register (same `machineID`) without pretending to be a heartbeat.
- Register handler calls `SetHostname` best-effort on the existing-node path when the reported hostname changed. Covers the edge case where credentials on disk expired and the node re-registers; heartbeats remain the primary refresh channel.
- Regression tests on the gorm store and the REST handler, both register and heartbeat paths, both "should update" and "must preserve when empty".
- Fakes in `server_test.go`, `middleware_test.go` and handlers `fakes_test.go` updated for the new signatures.

Agent-side change to send the hostname on every heartbeat is a separate kairos-agent PR.

Fixes kairos-io/kairos#4196

Signed-off-by: Itxaka <itxaka@kairos.io>

### DIFF
--- a/internal/store/gorm/adapters.go
+++ b/internal/store/gorm/adapters.go
@@ -33,8 +33,11 @@ func (a *NodeStoreAdapter) ListByLabels(ctx context.Context, labels map[string]s
 func (a *NodeStoreAdapter) ListBySelector(ctx context.Context, sel store.CommandSelector) ([]*store.ManagedNode, error) {
 	return a.S.ListBySelector(ctx, sel)
 }
-func (a *NodeStoreAdapter) UpdateHeartbeat(ctx context.Context, id string, agentVersion string, osRelease map[string]string, addresses []store.NodeAddress, bootState string) error {
-	return a.S.UpdateHeartbeat(ctx, id, agentVersion, osRelease, addresses, bootState)
+func (a *NodeStoreAdapter) UpdateHeartbeat(ctx context.Context, id string, agentVersion string, osRelease map[string]string, addresses []store.NodeAddress, bootState string, hostname string) error {
+	return a.S.UpdateHeartbeat(ctx, id, agentVersion, osRelease, addresses, bootState, hostname)
+}
+func (a *NodeStoreAdapter) SetHostname(ctx context.Context, id string, hostname string) error {
+	return a.S.SetHostname(ctx, id, hostname)
 }
 func (a *NodeStoreAdapter) UpdatePhase(ctx context.Context, id string, phase string) error {
 	return a.S.UpdatePhase(ctx, id, phase)
--- a/internal/store/gorm/store.go
+++ b/internal/store/gorm/store.go
@@ -253,7 +253,7 @@ func (s *Store) ListBySelector(ctx context.Context, sel store.CommandSelector) (
 	return nodes, nil
 }
 
-func (s *Store) UpdateHeartbeat(ctx context.Context, id string, agentVersion string, osRelease map[string]string, addresses []store.NodeAddress, bootState string) error {
+func (s *Store) UpdateHeartbeat(ctx context.Context, id string, agentVersion string, osRelease map[string]string, addresses []store.NodeAddress, bootState string, hostname string) error {
 	var n store.ManagedNode
 	if err := s.db.WithContext(ctx).First(&n, "id = ?", id).Error; err != nil {
 		return err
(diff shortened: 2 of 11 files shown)


Itxaka:
`e2e-tests` failed on infra flake in `docker/setup-qemu-action` step: `failed to set up container networking: bridge port not forwarding after 200ms`. Docker bridge issue on the runner, before any Go code ran. Will rerun once the parent workflow finishes (currently `test-bootable-raw-efi` still in progress).

PR #675, opened 2026-07-26
feat: atomic idempotent node claim/release for a group (kairos#4252)

wrkode:
## Summary

Adds an **atomic, idempotent node-claim endpoint** to the fleet API — the last hard plumbing blocker for building a Cluster API **infrastructure provider** on AuroraBoot (kairos-io/kairos#4252, epic kairos-io/kairos#120 / #4056).

Today the only ways to assign a group's node to something (`PUT /nodes/:id/group`, `PUT /nodes/:id/labels`) are last-write-wins with no compare-and-swap. That is unsafe for CAPI, whose reconcile is **level-triggered and retried**: a controller can crash mid-provision and reconcile the same machine again (must re-find *its own* node, not grab a second), and two machines reconciling concurrently can select the same node with no server-side arbitration.

## What's here

**`POST /api/v1/groups/:id/claim` `{ "claimKey": "<opaque caller key>" }`**
- Atomically assigns one **unclaimed** node in the group to `claimKey` and returns it.
- **Idempotent:** replaying with the same `claimKey` returns the **same** node, never a second — this is what makes a retried/restarted reconcile safe.
- Unknown group → `404` (told apart from capacity). Exhausted group → `409` with a machine-readable `code: "NoCapacity"` so the caller can surface "waiting for capacity" and requeue instead of failing.

**`POST /api/v1/nodes/:nodeID/release` `{ "claimKey": "..." }`**
- Returns the node to the pool for reuse, but only for the key that holds the claim.
- Already-unclaimed → `200 {released:false}` (idempotent). Held by a **different** key → `409 code:"ClaimMismatch"` so a caller can't drop someone else's claim.

The claim owner (`claimKey`/`claimedAt`) now surfaces in `GET /nodes` and `GET /nodes/:id`.

## Design

- `ManagedNode` gains `ClaimKey *string` + `ClaimedAt *time.Time`. `ClaimKey` is a **pointer** so "unclaimed" is SQL `NULL`: a composite **unique index** `idx_node_group_claim` on `(group_id, claim_key)` then makes a claimKey own **at most one** node per group, while `NULL`-exempt uniqueness lets many unclaimed nodes coexist — true on **both SQLite and PostgreSQL**.
- `NodeStore.ClaimNode` mirrors the store's existing compare-and-set idiom (`ClaimForDelivery`, `CASEjectState`): idempotent replay first, then a conditional `UPDATE ... WHERE id = ? AND claim_key IS NULL` over a bounded snapshot of unclaimed candidates. Exactly one of several concurrent callers wins a given node; a concurrent **same-key** claim is caught by the unique index (`ErrDuplicatedKey`, via newly-enabled gorm `TranslateError`) and resolved to the first winner. **No `FOR UPDATE` / `SKIP LOCKED`**, so it's correct on both drivers.
- Go client: `Groups.Claim`, `Nodes.Release`, and a `client.IsNoCapacity(err)` helper. OpenAPI spec regenerated via `make openapi` (additive only).

## Tests

Store-level (Ginkgo, file-backed WAL SQLite): idempotent replay, distinct-key isolation, no-capacity, cross-group isolation, release ownership, and two **concurrency invariants** — N distinct keys over M nodes yield exactly M single-winner claims plus N−M clean no-capacity results, and N concurrent same-key claims all return the one node. **Race-clean** under `-race`. Handler-level: claim success/idempotency/no-capacity/unknown-group/missing-key and release success/mismatch/no-op/missing-node/missing-key.

Backward compatible: the new columns default to `NULL`, and no existing behaviour changes.

Filed against kairos-io/kairos#4252 (issues are disabled on this repo).

### DIFF
--- a/docs/docs.go
+++ b/docs/docs.go
@@ -260,6 +260,63 @@ const docTemplate = `{
                 }
             }
         },
+        "/api/v1/artifacts/{id}/upload/{filename}": {
+            "put": {
+                "description": "Per-build endpoint the operator backend's exporter uses to ship finished artifacts back to AuroraBoot. Bearer is the UploadToken minted at Create time; the admin bearer does not grant access.",
+                "consumes": [
+                    "application/octet-stream"
+                ],
+                "tags": [
+                    "Artifacts"
+                ],
+                "summary": "Upload a single artifact file for a build",
+                "parameters": [
+                    {
+                        "type": "string",
+                        "description": "Artifact ID",
+                        "name": "id",
+                        "in": "path",
+                        "required": true
+                    },
+                    {
+                        "type": "string",
+                        "description": "Artifact filename (single-segment; no path separators)",
+                        "name": "filename",
+                        "in": "path",
+                        "required": true
+                    }
+                ],
+                "responses": {
+                    "201": {
+                        "description": "Created"
+                    },
+                    "400": {
+                        "description": "Bad Request",
+                        "schema": {
+                            "$ref": "#/definitions/handlers.APIError"
+                        }
+                    },
+                    "401": {
+                        "description": "Unauthorized",
+                        "schema": {
+                            "$ref": "#/definitions/handlers.APIError"
+                        }
+                    },
+                    "404": {
+                        "description": "Not Found",
+                        "schema": {
+                            "$ref": "#/definitions/handlers.APIError"
+                        }
+                    },
+                    "413": {
+                        "description": "Request Entity Too Large",
+                        "schema": {
+                            "$ref": "#/definitions/handlers.APIError"
+                        }
+                    }
+                }
+            }
+        },
         "/api/v1/groups": {
             "get": {
                 "security": [
@@ -439,6 +496,69 @@ const docTemplate = `{
                 "responses": {
                     "204": {
                         "description": "No Content"
+                    }
+                }
+            }
+        },
+        "/api/v1/groups/{id}/claim": {
+            "post": {
+                "security": [
+                    {
+                        "AdminBearer": []
+                    }
+                ],
+                "consumes": [
+                    "application/json"
+                ],
+                "produces": [
+                    "application/json"
+                ],
+                "tags": [
+                    "Groups"
+                ],
+                "summary": "Claim a node from a group",
+                "parameters": [
+                    {
+                        "type": "string",
+                        "description": "Group ID",
+                        "name": "id",
+                        "in": "path",
+                        "required": true
+                    },
+                    {
+                        "description": "Claim payload",
+                        "name": "body",
+                        "in": "body",
+                        "required": true,
+                        "schema": {
+                            "$ref": "#/definitions/handlers.APIClaimRequest"
+                        }
+                    }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "OK",
+                        "schema": {
+                            "$ref": "#/definitions/store.ManagedNode"
+                        }
+                    },
+                    "400": {
+                        "description": "Bad Request",
+                        "schema": {
+                            "$ref": "#/definitions/handlers.APIError"
+                        }
+                    },
+                    "404": {
+                        "description": "Not Found",
+                        "schema": {
+                            "$ref": "#/definitions/handlers.APIError"
+                        }
+                    },
+                    "409": {
+                        "description": "No unclaimed node available (code=NoCapacity)",
+                        "schema": {
+                            "$ref": "#/definitions/handlers.APIError"
+                        }
                     }
                 }
             }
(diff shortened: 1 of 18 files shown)


docs/docs.go
```diff
@@ -260,6 +260,63 @@ const docTemplate = `{
                 }
             }
         },
+        "/api/v1/artifacts/{id}/upload/{filename}": {
```
jimmykarily, line 263: [Off-topic] We keep forgetting to update this file (me this time). We need some gating in PRs to notify us when it needs to be updated.

internal/store/gorm/store.go
```diff
@@ -302,6 +307,105 @@ func (s *Store) NodeDelete(ctx context.Context, id string) error {
...
+		return nil, err
+	} else if mine != nil {
+		return mine, nil
+	}
+	return nil, store.ErrNoClaimCapacity
+}
+
+// ReleaseNode clears the claim on nodeID and returns it to the pool, but only if
+// the claim is currently held by claimKey. The conditional UPDATE (... WHERE id =
+// ? AND claim_key = ?) means a caller can only release its own claim: a node that
```
jimmykarily: Does this make claim_key a secret (needs to be protected) or just a way to ensure we claimers don't accidentally release each other's claims? Or to ask differently, is it here to prevent malicious acts? I think it's not, so we should soften the wording here from "a caller can only release its own claim" to "it ensures that callers only release their own claims" to avoid creating wrong assumptions.

jimmykarily: I think we need to clarify the roles of `id`, `claim_key`, `group_id`. As far as I understand, claim_key is what the claimer calls this machine. The naming logic resides on the claimer's side. `id` is the database auto-assigned id. group-id is a bit vague. Is it something like `my-nvidia-machines` or `amd64`/`arm64` etc?

jimmykarily (approved):
Code wise looks good to me. Whether it works as expected or not, you know better since you are consuming it for CAPI.

Itxaka:
@copilot resolve the merge conflicts in this pull request

internal/store/gorm/store.go
Itxaka: @jimmykarily is this a blocker? because you approved the PR so Im not sure if I should merge it/automerge it or wait for this to be resolved?

wrkode:
Thanks both. Addressed @jimmykarily's two doc points and rebased:

- **`ClaimNode` — identifier roles.** Spelled out in the doc comment: `groupID` is a **NodeGroup**, i.e. a *pool* of interchangeable nodes an operator has bucketed (e.g. by arch/capability like `arm64` or `gpu-nodes`) — the caller claims from a pool, not a named machine. `claimKey` is an **opaque, caller-owned label** for the claim (e.g. a CAPI Machine's UID); the naming logic lives entirely on the caller's side and it need not be secret. The returned node's own `id` is the store-assigned identifier of whichever node the pool handed back.
- **`ReleaseNode` — wording.** Softened "a caller can only release its own claim" → "ensures callers only release their own claims", and stated explicitly that the `claimKey` match is **accident-prevention, not a security boundary** (claimKey is readable via `GET /nodes`, so it's not an authorization control). You're right — it's exactly to stop claimers accidentally releasing each other's claims, not to prevent malicious acts.

Also rebased onto current `main` (clears the stale `e2e`/`riscv64` reds and picks up the otel-1.44 fix) and regenerated the OpenAPI spec.

@Itxaka — **not a blocker**, these were doc-only clarifications, now applied. Safe to automerge once CI is green. (The `scan-pr / osv-scan` red is the pre-existing repo-wide npm advisories — `react-router`/`brace-expansion` in the web UI — unrelated to this PR.)

internal/store/gorm/store.go
jimmykarily: not a blocker, but now fixed anyway

mudler:
@mudler-agent review

wrkode:
Small correction on my "once green" note: the two remaining red checks are both **repo-wide** and won't clear on this PR:

- **`build-linux-riscv64`** — broken since #683 (sbctl bump pulled in `go-piv/piv-go`, a CGO PIV/smartcard lib that fails to cross-compile for riscv64: `pkg-config libpcsclite not found`). It fails identically on every current open PR (#713, #709, #700). This PR adds no dependencies.
- **`scan-pr / osv-scan`** — the pre-existing npm advisories (partly addressed by #684), also red repo-wide.

So `unit-tests` / `e2e-tests` / `build-image` / bootable suites all pass here — the node-claim change itself is green. The riscv red is a separate repo-CI issue (probably wants `libpcsclite` added to the riscv cross-build, or `go-piv` build-tagged out of the riscv target). Happy to open a separate PR for that if useful — just didn't want to fold an unrelated CI fix into this one. Merge is your call (admin-merge vs waiting on the repo-wide riscv fix).

mudler-agent (approved):
Reviewed the store CAS and unique-index arbitration, handler/client contract, release semantics, generated API surface, and concurrency coverage. The same-key and distinct-key race cases are covered, and the targeted GORM store tests pass locally. No blocking findings.

PR #680, opened 2026-07-28
fix(raw-disk): make recovery image size configurable

### LINKED ISSUE #4230: Raw/VHD disk build fails for large airgap payloads because COS_SYSTEM sizing is too tight
**Kairos version:**

Auroraboot v0.25.0

**CPU architecture, OS, and Version:**

**Describe the bug**

AuroraBoot fails when building a VHD/raw disk image from a Kairos image that includes a large airgap payload under /data.

  The failure happens while AuroraBoot creates the internal COS_SYSTEM recovery image. The generated ext2 image is too small, and rsync fails with No space left on device.

  The final disk size can be configured with disk.size, and the state partition can be configured with disk.state_size, but this failure happens before the final raw/VHD disk is assembled. As far as I can tell, the size of this internal COS_SYSTEM
  image is not currently configurable.

  This is important for airgapped deployments where OCI image tarballs need to be embedded into the image and loaded into a local registry on first boot.

**To Reproduce**
Build a VHD with AuroraBoot using a container image that contains several GiB of airgap assets under /data.

  Command used:
```
  auroraboot \
    --debug=true \
    --set "disable_http_server=true" \
    --set "disable_netboot=true" \
    --set "state_dir=/build" \
    --set "disk.vhd=true" \
    --set "disk.size=65000" \
    --set "disk.state_size=30000" \
    --set "container_image=${CONTAINER_IMAGE}" \
    --cloud-config /config/cloud-config.yaml
```
  The extracted rootfs during the build is around 7.6 GiB:

  du -sm /build/temp-rootfs
  7618    /build/temp-rootfs

  du -sm /build/temp-rootfs/data
  5746    /build/temp-rootfs/data

  find /build/temp-rootfs -xdev -printf '.' | wc -c
  78839

  Most of /data contains OCI image tarballs for an airgapped install. This is intentional: the target deployment has no access to external registries on first boot.

**Expected behavior**

AuroraBoot should successfully build the raw/VHD disk image, or provide a supported way to configure the size of the internal COS_SYSTEM / recovery system image.

**Logs**
```
 rsync: [receiver] write failed on "/build/tmp/auroraboot-raw-image-219422212/recovery-img/var/lib/rpm/rpmdb.sqlite": No space left on device (28)
  rsync error: error in file IO (code 11) at receiver.c(381) [receiver=3.4.1]

  rsync: [sender] write error: Broken pipe (32)

  2026-07-09T04:42:26Z DBG [10] Unmounting image COS_SYSTEM
  2026-07-09T04:42:40Z DBG Opening loop device device=/dev/loop6
  2026-07-09T04:42:40Z DBG [10] Clearing loop device
  2026-07-09T04:42:40Z ERR failed to create recovery image error="exit status 11" image={"fs":"ext2","label":"COS_SYSTEM","size":7974,"source":{}} source=/build/temp-rootfs
  2026-07-09T04:42:41Z ERR failed to create recovery partition error="exit status 11"
  2026-07-09T04:42:41Z ERR Generating raw disk '/build' from '/build/temp-rootfs' failed with error 'exit status 11'

  2026-07-09T04:42:44Z DBG 3.
  2026-07-09T04:42:44Z DBG  <gen-raw-efi-disk> (error: 1 error occurred:
          * exit status 11

  ) (background: false)

  2026-07-09T04:42:44Z DBG 4.
  2026-07-09T04:42:44Z DBG  <convert-vhd> (error: 'convert-vhd' deps gen-raw-efi-disk failed) (background: false)

  2 errors occurred:
          * exit status 11
          * 'convert-vhd' deps gen-raw-efi-disk failed
          *
```

**Additional context**
(DISCLAIMER: Analyzed the code with the help of AI since I'm not a go developer)
 From reading the AuroraBoot code, the internal COS_SYSTEM size appears to be calculated around here:

  size, _ := config.GetSourceSize(r.config, recoveryImage.Source)
  recoveryImage.Size = uint(size + 200)

  This produces an image of about 7.9 GiB in my case:

  image={"fs":"ext2","label":"COS_SYSTEM","size":7974}

  But the copy still fails with No space left on device.

  This is not caused by the final VHD size. The build fails before VHD conversion starts

mudler-agent:
## Summary

- add `disk.recovery_image_size` for EFI, BIOS, raw, GCE, and VHD generation
- preserve automatic source-size headroom when the override is unset or non-positive
- reject configured recovery images smaller than the measured source before deployment
- size the recovery partition from the selected recovery image so large overrides fit

## Testing

- `go test ./pkg/ops ./deployer`
- `go vet ./pkg/ops ./deployer`

The repository-wide `go test ./...` cannot complete in this environment because uncached module downloads return HTTP 403 and generated UI assets are absent.

Fixes kairos-io/kairos#4230

### DIFF
--- a/pkg/ops/rawDiskGeneration.go
+++ b/pkg/ops/rawDiskGeneration.go
@@ -47,7 +47,7 @@ type RawImage struct {
 	Output               string               // Output image destination dir. Final image name will be based on the contents of the source /etc/kairos-release file
 	FinalSize            uint64               // Final size of the disk image in MB
 	StateSize            int64                // Size of the state partition in MB
-	SystemSize           int64                // Size of the system image in MB
+	RecoveryImageSize    int64                // Size of the recovery image in MB
 	tmpDir               string               // A temp dir to do all work on
 	elemental            *elemental.Elemental // Elemental instance to use for the operations
 	efi                  bool                 // If the image should be EFI or BIOS
@@ -62,7 +62,7 @@ type RawImage struct {
 
 // NewEFIRawImage creates a new RawImage struct
 // config is initialized with a default config to use the standard logger
-func NewEFIRawImage(source, output, cc string, finalsize uint64, stateSize, systemSize int64, noDefaultCloudConfig bool) *RawImage {
+func NewEFIRawImage(source, output, cc string, finalsize uint64, stateSize, recoveryImageSize int64, noDefaultCloudConfig bool) *RawImage {
 	cfg := config.NewConfig(config.WithLogger(internal.Log))
 	return &RawImage{
 		efi:                  true,
@@ -73,12 +73,12 @@ func NewEFIRawImage(source, output, cc string, finalsize uint64, stateSize, syst
 		CloudConfig:          cc,
 		FinalSize:            finalsize,
 		StateSize:            stateSize,
-		SystemSize:           systemSize,
+		RecoveryImageSize:    recoveryImageSize,
 		NoDefaultCloudConfig: noDefaultCloudConfig,
 	}
 }
 
-func NewBiosRawImage(source, output string, cc string, finalsize uint64, stateSize, systemSize int64, noDefaultCloudConfig bool) *RawImage {
+func NewBiosRawImage(source, output string, cc string, finalsize uint64, stateSize, recoveryImageSize int64, noDefaultCloudConfig bool) *RawImage {
 	cfg := config.NewConfig(config.WithLogger(internal.Log))
 	return &RawImage{efi: false,
 		config:               cfg,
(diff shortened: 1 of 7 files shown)


pkg/ops/rawDiskGeneration.go
```diff
@@ -339,8 +342,10 @@ func (r *RawImage) createRecoveryPartitionImage() (string, error) {
 		MountPoint: tmpDirRecoveryImage,
 	}
 	size, _ := config.GetSourceSize(r.config, recoveryImage.Source)
-	// Add some extra space to the image in case the calculation is a bit off
-	recoveryImage.Size = uint(size + 200)
+	recoveryImage.Size = systemImageSize(uint64(size), r.SystemSize)
```
Itxaka: add a check if sourcesize is bigger than systemsize, fail hard ? Otherwise it will just fail later when copying stuff into the recovery image.

Itxaka (changes requested):
please rename the vars. Its not the system image, its the recovery mage so make it clear. Recovery image is stored in teh system partition yes, but its not called system image anywhere.

Itxaka (approved):

PR #685, opened 2026-07-28
feat: add a boot-to-RAM submenu to the live menu

Itxaka:
Adds a live grub entry for immucore's kairos.ram workflow: the OS runs
from memory while immucore mounts COS_OEM + COS_PERSISTENT from the local
disk, creating them on first boot if the disk is empty
(kairos.ram.create_partitions).

Layout:

- "Kairos (boot to RAM)" is the last of the normal live entries
- it is a submenu, so selecting it boots nothing
- it opens with the caveats — new partition table on an empty disk, and
  no users in the image, so a cloud config defining one has to reach the
  system somehow (bundled, datasource, or config_url)
- the last line, "Boot Kairos to RAM now", is what actually boots
- ESC goes back to the main menu

Since this path can write a fresh partition table to the system disk,
having to step through the warnings to get to it is deliberate.

The caveats are rendered as menuentries with no-op bodies. That looks odd
but is the only thing that works everywhere. A previous version echoed
the text and paused on `sleep --verbose --interruptible 30`; that worked
on BIOS only, because we control eltorito.img and had rebuilt it with the
sleep module, while grub.efi comes from the rootfs distro and may not
ship it — Hadron's does not, giving "can't find command `sleep`" on EFI.
Echo without a pause is useless too, grub clears the screen the moment a
menuentry body returns without booting. Menu lines need nothing beyond
normal.mod, and the patched eltorito.img can go away.

### DIFF
--- a/pkg/constants/grub_live_bios.cfg
+++ b/pkg/constants/grub_live_bios.cfg
@@ -61,6 +61,32 @@ menuentry "Kairos (debug)" --class os --unrestricted {
     initrd ($root)/boot/initrd
 }
 
+# The in-RAM workflow (kairos.ram) is only useful when a cloud config with at
+# least one user reaches the system, and it writes a partition table to an
+# empty disk on first boot (kairos.ram.create_partitions), so it hides behind
+# a submenu that explains all that before booting. The warning is rendered as
+# informational menu lines (no-op bodies) because that is the only way to keep
+# text on screen that works on every grub: echo output is cleared when the
+# menu redraws, and pausing needs the sleep/read modules, which are not
+# guaranteed to be built into the EFI grub image taken from the rootfs.
+# The last entry is the one that actually boots; ESC goes back.
+submenu "Kairos (boot to RAM)" --class os --unrestricted {
+    menuentry "'Boot to RAM' runs the OS entirely from memory, nothing is installed" --unrestricted { set _info=1 }
+    menuentry "immucore mounts COS_OEM and COS_PERSISTENT from the system disk; on an" --unrestricted { set _info=1 }
+    menuentry "empty disk it CREATES them, writing a new partition table on first boot" --unrestricted { set _info=1 }
+    menuentry "The image ships with no users: unless a cloud config defining at least" --unrestricted { set _info=1 }
+    menuentry "one user reaches the system, you will not be able to log in. Provide it" --unrestricted { set _info=1 }
+    menuentry "via a config bundled in the image, a datasource (user-data on a cloud" --unrestricted { set _info=1 }
+    menuentry "provider or attached cdrom/usb), or a remote config_url stanza" --unrestricted { set _info=1 }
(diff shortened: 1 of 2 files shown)


pkg/constants/grub_live_bios.cfg
```diff
@@ -54,6 +54,17 @@ menuentry "Kairos (boot local node from livecd)" --class os --unrestricted {
     initrd ($root)/boot/initrd
 }
 
+# In-RAM workflow (kairos.ram): immucore mounts the local COS_OEM +
+# COS_PERSISTENT partitions from disk, auto-creating them on first boot when
+# the disk is empty. No install-mode and no rd.cos.disable: immucore must run
+# its full in-RAM DAG here.
+menuentry "Kairos (boot to RAM)" --class os --unrestricted {
```
Itxaka: @copilot add a more explicit title as this will vreate partitions in any disk that it finds that its empty on selection so its gotta be careful when choosing it

Itxaka:
Tested end to end in QEMU with an ISO built from `quay.io/kairos/alpine:3.21-core-amd64-generic-v3.6.0`.

BIOS (SeaBIOS) needed the `eltorito.img` rebuild — the embedded grub core image had no `sleep` module, so grub printed `error: can't find command 'sleep'` and booted immediately with no countdown. With the rebuilt image both firmwares behave the same: `Advanced boot settings` shows up at the bottom of the live menu, the entry prints the create_partitions and no-users explanation, the 30s countdown ticks down, ESC interrupts it and returns to the menu without booting, and letting it run out boots the kernel with `kairos.ram kairos.ram.create_partitions` on the cmdline (confirmed on serial).

EFI (OVMF) was already fine before the rebuild since that grub comes from the rootfs.

jimmykarily (approved):

Itxaka:
Re-verified the reworked menu in QEMU on both firmware paths, alpine 3.21 core base:

- BIOS (SeaBIOS): main menu -> Advanced boot settings -> Kairos (boot to RAM) renders the info lines + "Boot Kairos to RAM now", ESC backs out, booting the last entry gets `kairos.ram kairos.ram.create_partitions` on the kernel cmdline (serial log).
- EFI (OVMF): same flow, same cmdline. No sleep module involved anymore so no missing-command failure regardless of what grub.efi the rootfs ships.

Unrelated thing spotted while testing: on EFI the Thor smbios detection can leak `$thor_options` into the cmdline on plain QEMU (`test $model == "Thor"` misbehaves when the model string is empty/multi-word). Pre-existing, will look at it separately.

Itxaka:
Re-verified after flattening the layout (submenu now sits at the same level as the other live entries): BIOS and EFI QEMU boots both navigate the submenu, ESC returns to the main menu (highlight resets to the first entry, grub behavior), and booting the last line lands `kairos.ram kairos.ram.create_partitions` on the kernel cmdline on both firmware paths.

PR #686, opened 2026-07-28
fix(uki): enable initramfs networking

### LINKED ISSUE #3758: Fix early boot networking in UKI mode
In UKI mode, after installation, there is no proper network setup and the KMS url cannot be resolved (DNS issue) and can't be reached in general (general networking issue, no interface is up etc).

This is the only thing left to have remote KMS working in UKI mode after finishing this story: https://github.com/kairos-io/kairos/issues/3723

mudler-agent:
## Summary

- add `rd.neednet=1` to AuroraBoot UKI kernel command lines
- limit forced initramfs networking to UKI images instead of all dracut-based Kairos images
- add regression coverage for the UKI command line

## Tests

- `GOPROXY=direct go test ./pkg/constants ./pkg/utils`
- `git diff --check`

`go test ./pkg/uki` could not start locally because fetching the newly bumped OpenTelemetry modules returns HTTP 403 in this environment.

Fixes kairos-io/kairos#3758

Supersedes kairos-io/kairos-init#414.

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -82,7 +82,7 @@ const (
 	Archaarch64 = "aarch64"
 	ArchRiscv64 = "riscv64"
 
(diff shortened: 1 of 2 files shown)


Itxaka (changes requested):
You didnt understand the purpose or ask.of the original ticket. Close this please

PR #711, opened 2026-08-04
fix(iso): make live ISOs boot on strict UEFI firmware

jimmykarily:
Two independent defects were preventing the live ISO from booting on some UEFI stacks (VMware ESXi/Workstation, some OVMF/BMC virtual media). Both manifest as "no bootable device" or a silent drop through the boot manager, so the reporter hit them together.

1. MBR-only ISO xorriso was invoked without appended_part_as=gpt, so the ESP was only exposed via MBR partition entries. Strict firmware enumerates ESPs via GPT only and refuses to boot such ISOs. Add the flag so xorriso emits a GPT header alongside the protective MBR, matching Ubuntu's layout.

2. Wrong grub EFI binary The rootfs search list preferred grubx64.efi.signed (the disk-installer grub whose prefix is /EFI/ubuntu and which lacks iso9660 baked in). On firmware that sets \$root to the ISO9660 device, that grub cannot read its own filesystem, let alone find its config, and drops to rescue. Prefer gcdx64.efi.signed / gcdaa64.efi.signed for live media, which have iso9660 baked in and prefix /boot/grub.

   To keep shim chainloading working under Secure Boot, cleanupGrubName now renames gcd*.efi to grub*.efi on copy (shim looks for grubx64.efi next to itself). A grub.cfg is also written at /boot/grub/grub.cfg (both in the EFI image for the USB case and on the ISO root for the CD case) so the CD grub finds its config regardless of what the firmware set \$root to.

### DIFF
--- a/pkg/ops/iso.go
+++ b/pkg/ops/iso.go
@@ -461,6 +461,18 @@ func (b BuildISOAction) createEFI(rootdir string, isoDir string) error {
 		b.cfg.Logger.Errorf("Failed writing grub.cfg: %v", err)
 		return err
 	}
+	// The CD grub (gcd*.efi.signed) has its prefix baked to /boot/grub, so
+	// also drop a grub.cfg there. Placed both in the EFI image (USB case,
+	// where $root is the FAT ESP) and in the ISO root (CD case, where $root
+	// is the ISO9660 device).
+	if err = b.writeCdGrubEfiCfg(temp); err != nil {
+		b.cfg.Logger.Errorf("Failed writing CD grub.cfg: %v", err)
+		return err
+	}
+	if err = b.writeCdGrubEfiCfg(isoDir); err != nil {
+		b.cfg.Logger.Errorf("Failed writing CD grub.cfg: %v", err)
+		return err
+	}
 
 	// For RISC-V, create startup.nsh to auto-boot since some UEFI implementations
 	// don't automatically load the fallback boot path (BOOTRISCV64.EFI)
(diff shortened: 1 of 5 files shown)


pkg/ops/iso.go
```diff
@@ -642,16 +667,26 @@ func (b BuildISOAction) copyShim(tempdir, rootdir string) error {
...
-		return append(
-			[]string{"/usr/share/efi/riscv64/grub.efi"},
-			sdkutils.GetEfiGrubFiles(arch)...,
-		)
-	}
-	return sdkutils.GetEfiGrubFiles(arch)
+	sdkPaths := sdkutils.GetEfiGrubFiles(arch)
+	switch {
+	case utils.IsRiscv64(arch):
+		return append([]string{"/usr/share/efi/riscv64/grub.efi"}, sdkPaths...)
```
mudler, line 683: I think this is better that belongs to the sdk. We can have a specific entry for LiveCD EFI files to search

mudler:
@mudler-agent review

mudler-agent (commented):
I reviewed the full diff and agree with the existing SDK ownership comment. The live-media path ordering should be exposed by kairos-sdk as a separate helper, for example `GetEfiLiveGrubFiles`, while preserving `GetEfiGrubFiles` for raw-disk consumers. AuroraBoot can then use the live helper here without duplicating distro-specific GRUB paths. I found no additional correctness issue in the GPT publication, GRUB filename normalization, or config placement changes. The constants tests pass locally; the ops test could not run locally because this environment receives 403s from proxy.golang.org, while the PR unit-test job is green.

mudler:
@mudler-agent can you do the changes on the SDK and open up a PR there?

mudler:
@mudler-agent pick this up now on a new PR and base it on the commit from the SDK repo

mudler-agent:
This PR is now superseded by #713, which carries the strict-UEFI fix using the SDK helper from kairos-io/kairos-sdk#823. Both replacements merged on August 6 and August 5 respectively, and current `main` contains the fix as `2f1ecb9`. This older draft is conflicting and can be closed.

PR #713, opened 2026-08-05
fix(iso): make live ISOs boot on strict UEFI firmware

mudler-agent:
## Summary

- publish the appended ESP through GPT for strict UEFI firmware
- prefer live-media GRUB binaries through kairos-sdk `GetEfiLiveGrubFiles`
- place GRUB configuration at the CD-media prefix and normalize shim payload names

This carries the fix from #711 into a bot-owned follow-up and bases the GRUB path selection on kairos-sdk#823 as requested. The SDK pin also includes kairos-sdk#822.

## Testing

- The source PR unit tests were green.
- Local `go test ./pkg/constants ./pkg/ops` is currently blocked because this environment receives HTTP 403 for `go.opentelemetry.io/otel` from both proxy.golang.org and goproxy.io.

Signed-off-by: Ettore Di Giacinto <mudler@localai.io>


pkg/constants/constants.go
```diff
@@ -141,6 +146,12 @@ func GetXorrisoBooloaderArgs(root string) []string {
 		"-boot_image", "any", "emul_type=no_emulation",
 		"-append_partition", "2", "0xef", filepath.Join(root, IsoEFIPath),
 		"-boot_image", "any", "efi_path=--interval:appended_partition_2:all::",
+		// Publish appended partitions in a GPT header (with a protective MBR)
+		// so strict UEFI firmwares (VMware ESXi, some OVMF/BMC stacks) that
+		// enumerate the ESP via GPT can find and boot the ISO. Without this
+		// the ISO carries an MBR-only partition table and boot fails on
+		// those firmwares.
+		"-boot_image", "any", "appended_part_as=gpt",
```
mudler, line 154: this changes the default from MBR. While I don't think should be an issue, we might want to have a flag/arg to turn this off somehow until is proven to be good as default

https://man.archlinux.org/man/xorriso.1.en

mudler: maybe we should use gpt_disk_guid here instead? @mudler-agent double check

mudler: Poked at the MBR concern. @mudler-agent try to build the ISO both with and without appended_part_as=gpt and diff the partition tables. The MBR boot code (first 446 bytes, what BIOS actually runs) should come out identical, and isohybrid boots via the El Torito boot info anyway, not the partition entries. So BIOS boot should still work. It's also the same protective MBR + GPT layout Debian/Ubuntu/Fedora ship, so I'm not too worried about the common case.

One thing to watch: the protective MBR loses the active/bootable flag we used to have. A few picky BIOSes refuse to boot without one (the old xorriso #46716 thing). Narrow and legacy only.

Easy fix is adding mbr_force_bootable=on. @mudler-agent you should test it if it puts the bootable flag back while keeping the GPT.  IF it does, add that, and still fine keeping this behind a flag until we've proven it as default.

gpt_disk_guid isn't a boot thing, it's mostly for reproducible builds. Worth setting but not a blocker.

PR #716, opened 2026-08-06
fix(build-iso): apply --overlay-iso via StepInjectCC + run xorriso when only overlay is present

### LINKED ISSUE #4283: AuroraBoot: `build-iso` subcommand silently drops `--overlay-iso` content (StepInjectCC not called)
## Summary

`auroraboot build-iso <source> --overlay-iso <dir>` on the CLI produces an ISO that does **not** contain any of the files placed in `<dir>`. No error, no warning — the overlay is silently discarded.

The equivalent pipeline invocation (`auroraboot --set container_image=... --set iso.overlay_iso=... --cloud-config ...`) *does* apply the overlay correctly.

Affects: **v0.26.2** (verified) and every earlier version I could inspect (v0.26.1, v0.18.0).

## Reproducer

Any Ubuntu-based rootfs works; contents don't matter. Minimal case:

```bash
# Prepare an overlay dir with a canary file
mkdir -p /tmp/overlay/canary
echo "hello from overlay" > /tmp/overlay/canary/marker.txt

# Build an ISO with --overlay-iso pointing at it
auroraboot build-iso dir:/path/to/kairos-rootfs \
    --overlay-iso /tmp/overlay \
    --output /tmp/out \
    --arch amd64

# Look for the canary on the resulting ISO
xorriso -indev /tmp/out/*.iso -osirrox on \
        -extract /canary/marker.txt /tmp/extracted.txt 2>&1 | grep -i 'error\|not found'
# → xorriso : FAILURE : Cannot find in ISO image: '/canary/marker.txt'
```

The `/canary/marker.txt` file is nowhere on the ISO tree. Same overlay via pipeline mode:

```bash
auroraboot --set container_image=<image> \
           --set iso.overlay_iso=/tmp/overlay \
           --set disable_http_server=true \
           --set disable_netboot=true \
           --cloud-config /some/cloud-init.yaml \
           --set state_dir=/tmp/aurora

xorriso -indev /tmp/aurora/*.iso -osirrox on \
        -extract /canary/marker.txt /tmp/extracted.txt
# → succeeds; file is on the ISO
```

## Root cause

`internal/cmd/build-iso.go` (v0.26.2) at [lines 144-149](https://github.com/kairos-io/AuroraBoot/blob/v0.26.2/internal/cmd/build-iso.go#L144-L149) drives the subcommand through this step list:

```go
for _, step := range []func() error{
    d.PrepDirs,
    d.StepCopyCloudConfig,
    d.StepDumpSource,
    d.StepGenISO,
} {
    ...
}
```

**`StepInjectCC` is not in the list.** That step (defined in [`deployer/steps.go` `StepInjectCC()`](https://github.com/kairos-io/AuroraBoot/blob/v0.26.2/deployer/steps.go)) is what runs `ops.InjectISO`, which contains the `xorriso -map <overlay> / -boot_image any replay` command that actually copies overlay content onto the finalised ISO tree.

Pipeline mode's `Deployer.Run()` includes `StepInjectCC` in its step graph, which is why pipeline mode works.

`StepGenISO` on its own *does* wire the overlay into `LiveISO.Image` via `append(spec.Image, imagetypes.NewDirSrc(i.OverlayISO))` at [`pkg/ops/iso.go:167-168`](https://github.com/kairos-io/AuroraBoot/blob/v0.26.2/pkg/ops/iso.go#L167-L168), but empirically that path doesn't reliably put overlay files onto the finalised ISO — `prepareBootArtifacts` runs afterwards and `/boot/grub2/grub.cfg` is the only path guarded by an "already exists" check ([line 371-373](https://github.com/kairos-io/AuroraBoot/blob/v0.26.2/pkg/ops/iso.go#L371-L373)).

## Impact

Any downstream project using the documented `auroraboot build-iso ... --overlay-iso ...` CLI is silently building ISOs missing:
- Custom `/boot/grub2/grub.cfg` (menu branding, kernel cmdline customisation)
- Cloud-config / user-data embedded in the ISO
- Cluster config bundles (SPC), content bundles, custom scripts, etc.

CanvOS (Spectro Palette Edge) hit this while migrating from `osbuilder-tools` to AuroraBoot v0.26.x — we lost the entire Palette-branded installer menu, and every custom file we placed under `overlay/files-iso/` disappeared from the ISO. We currently work around it in our Earthfile with our own `xorriso -indev X -outdev X -map /overlay / -boot_image any replay` step after the AuroraBoot build.

## Suggested fix

Two options, in decreasing preference:

1. **Add `d.StepInjectCC` to the subcommand's step list** in `internal/cmd/build-iso.go`:
   ```diff
    for _, step := range []func() error{
        d.PrepDirs,
        d.StepCopyCloudConfig,
        d.StepDumpSource,
        d.StepGenISO,
   +    d.StepInjectCC,
    } {
   ```
   `StepInjectCC` already guards on `!d.rawDiskIsSet() && d.isoOption()` so it's a no-op for non-ISO builds. This is a one-line fix.

2. **Or:** Make `StepGenISO` reliably apply overlay end-to-end without needing InjectISO afterwards. Would require reordering `applySources` and `prepareBootArtifacts` so overlay wins, and probably guarding more files against overwrite (currently only `/boot/grub2/grub.cfg` has the "skip if exists" check).

Option 1 is a diff of one identifier — trivial and matches pipeline mode exactly.

I'm preparing a PR with the one-line fix plus a companion fix for a related gotcha in `InjectISO` (see below), and a regression test.

## Companion problem worth calling out

`ops.InjectISO` at [`pkg/ops/iso.go:180-227`](https://github.com/kairos-io/AuroraBoot/blob/v0.26.2/pkg/ops/iso.go#L180-L227) gates the `xorriso -map <overlay> /` invocation on the existence of a non-empty `<state>/config.yaml`:

```go
if i.OverlayISO != "" {
    copy.Copy(i.OverlayISO, tmp)   // copies overlay to tmp
}
if _, err := os.Stat(filepath.Join(dst, "config.yaml")); err == nil {
    f, _ := os.ReadFile(...)
    if f != nil && len(f) > 0 {
        copy.Copy(dst/config.yaml, tmp/config.yaml)
        sdkutils.SH("xorriso ... -map tmp / -boot_image any replay")  // ← only fires here
    }
}
```

So callers passing `--overlay-iso` **without** a `--cloud-config` (or with an empty one) will still see the overlay silently dropped even after Option 1 above. The fix is to run xorriso whenever `OverlayISO != ""` OR a non-empty cloud-config is present. Planning to bundle this into the same PR.

## Related

Filing a separate issue on `--output` and `--override-name` being inert for `dir:` sources (they're not read on the subcommand path either).

Downstream that hit this: CanvOS (Spectro Palette Edge), migrating from `osbuilder-tools` to AuroraBoot v0.26.x.

vipsharm:
## What

Fixes the `build-iso` subcommand so `--overlay-iso` actually applies overlay content to the produced ISO. Also makes `InjectISO` run the xorriso injection when only overlay (no cloud-config) is present.

## Why

The `build-iso` subcommand currently runs:

```go
for _, step := range []func() error{
    d.PrepDirs,
    d.StepCopyCloudConfig,
    d.StepDumpSource,
    d.StepGenISO,
} {
    ...
}
```

`StepInjectCC` (which wraps `ops.InjectISO` — the `xorriso -map <overlay> / -boot_image any replay` step that puts overlay content onto the finalised ISO) is **not in the list**. Pipeline mode's `Deployer.Run()` has it, which is why `--set iso.overlay_iso=...` works there but `--overlay-iso` does not work via the subcommand.

Effect for downstream users: every file placed under `--overlay-iso <dir>` is silently discarded — no error, no warning. Custom grub menu, cloud-config, cluster bundles, content bundles, custom scripts, all lost.

Additionally, even with `StepInjectCC` wired in, `ops.InjectISO` gated the `xorriso -map` call inside the "non-empty `config.yaml` in `dst`" branch, so a caller passing `--overlay-iso` but no `--cloud-config` still lost the overlay.

## Changes

**`internal/cmd/build-iso.go`** — one-line addition to the step list:

```diff
     d.PrepDirs,
     d.StepCopyCloudConfig,
     d.StepDumpSource,
     d.StepGenISO,
+    d.StepInjectCC,
```

`StepInjectCC` already guards on `!d.rawDiskIsSet() && d.isoOption()`, so this is a no-op for non-ISO builds and matches pipeline-mode behavior exactly.

**`pkg/ops/iso.go`** — refactor `InjectISO` so the `xorriso -map` runs whenever `OverlayISO != ""` OR a non-empty cloud-config is present, not only when a non-empty cloud-config is present. Tracked with a single `haveContent` flag. No behavior change for callers that had a cloud-config (the pre-existing test still passes unchanged).

**`pkg/ops/inject_iso_test.go`** — new regression test that exercises the overlay-only branch (empty `dst`, `schema.ISO{OverlayISO: overlay}` with a canary file), and asserts the canary file lands on the produced ISO.

## Manual repro

**Before this PR:**

```bash
mkdir -p /tmp/overlay/canary
echo "canary" > /tmp/overlay/canary/marker.txt

auroraboot build-iso dir:/path/to/rootfs \
    --overlay-iso /tmp/overlay \
    --output /tmp/out \
    --arch amd64

xorriso -indev /tmp/out/*.iso -osirrox on \
    -extract /canary/marker.txt /tmp/extracted 2>&1 | grep FAILURE
# → xorriso : FAILURE : Cannot find in ISO image: '/canary/marker.txt'
```

**After this PR:**

```bash
# (same commands)
# → /tmp/extracted exists and contains "canary"
```

## Downstream context

CanvOS (Spectro Palette Edge) hit this while migrating from `osbuilder-tools` to AuroraBoot v0.26.x — the entire Palette-branded installer menu and every file under `overlay/files-iso/` (user-data, cluster config, content bundles) silently disappeared from the ISO. Our downstream workaround was a `xorriso -indev X -outdev X -map /overlay / -boot_image any replay` step after `auroraboot build-iso`, which is exactly what `StepInjectCC` does — so wiring it into the subcommand path lets us drop that workaround.

## Fixes

Fixes kairos-io/kairos#4283

Related: kairos-io/kairos#4284 (`--output` / `--override-name` inert for `dir:` sources on the same subcommand path — worth a separate look).

## Test plan

- [ ] `go build ./...` (cross-compiled for Linux locally, passes)
- [ ] `go vet ./pkg/ops/... ./internal/cmd/...` (passes)
- [ ] `ginkgo ./pkg/ops/` with `xorriso` installed — the new regression test + the existing cloud-config test both pass
- [ ] Manual: `auroraboot build-iso dir:X --overlay-iso Y --arch amd64` produces an ISO containing Y's contents
- [ ] Manual regression: same command with a non-empty `--cloud-config` still injects the cloud config (unchanged behavior)


pkg/ops/iso.go
```diff
@@ -188,41 +188,55 @@ func InjectISO(dstFunc, isoFunc valueGetOnCall, i schema.ISO) func(ctx context.C
...
-		return err
+		if !haveContent {
+			return nil
+		}
+
+		// No os.Chdir here: every path below (isoFile, tmp, and the
+		// config.yaml source) is already absolute, and xorriso is given
+		// absolute -indev/-outdev/-map paths. Builds run as concurrent
+		// goroutines in `web` mode, where a process-global chdir would
+		// corrupt sibling builds — so we must not mutate the cwd.
```
jimmykarily, line 231: No need to describe why we don't do what we don't do. We just don't do it. I know this comment existed before but it's useless anyway.

internal/cmd/build-iso.go
```diff
@@ -146,6 +146,11 @@ var BuildISOCmd = cli.Command{
 			d.StepCopyCloudConfig,
 			d.StepDumpSource,
 			d.StepGenISO,
+			// StepInjectCC is what actually applies --overlay-iso (and the
+			// cloud-config) onto the finalised ISO tree via xorriso. Without
+			// it the subcommand silently drops overlay content that pipeline
+			// mode (Deployer.Run) applies correctly. See #4283.
```
jimmykarily, line 152:
```suggestion
			// mode (Deployer.Run) applies correctly. See kairos#4283.
```

jimmykarily (approved):

jimmykarily:
I had Claude digging in and it seems that the `internal/cmd/build-iso.go` half of this PR is a no-op. `StepInjectCC` at `deployer/steps.go:183-190` has

```go
herd.EnableIf(func() bool { return !d.rawDiskIsSet() && d.isoOption() }),
```

and `isoOption()` at `deployer/steps.go:307-309` is `return !d.fromImage()`. In the subcommand, `ReleaseArtifact.ContainerImage` is always set from `<source>` (any source shape: `oci:`, `docker://`, `dir:`), so `fromImage()` is always true, `isoOption()` is always false, `EnableIf` returns false, and herd marks the step `Ignored=true`. `herd/dag.go:122` then skips the callback during `Run`. So adding `d.StepInjectCC` to the step list registers a step that will not execute.

Here is a self-contained test that demonstrates this. Save it as `deployer/inject_cc_wiring_test.go` on top of this PR and run `go test ./deployer/ -run TestDeployer -v -args -ginkgo.focus="StepInjectCC herd wiring"`. Both specs pass on the current PR head:

```go
package deployer_test

import (
	"github.com/kairos-io/AuroraBoot/deployer"
	"github.com/kairos-io/AuroraBoot/pkg/constants"
	"github.com/kairos-io/AuroraBoot/pkg/schema"
	"github.com/spectrocloud-labs/herd"

	. "github.com/onsi/ginkgo/v2"
	. "github.com/onsi/gomega"
)

func findEntry(d *deployer.Deployer, name string) *herd.GraphEntry {
	for _, layer := range d.Analyze() {
		for i, op := range layer {
			if op.Name == name {
				return &layer[i]
			}
		}
	}
	return nil
}

var _ = Describe("StepInjectCC herd wiring on the build-iso subcommand path", Label("wiring"), func() {
	registerBuildISOSteps := func(d *deployer.Deployer) {
		for _, step := range []func() error{
			d.PrepDirs,
			d.StepCopyCloudConfig,
			d.StepDumpSource,
			d.StepGenISO,
			d.StepInjectCC,
		} {
			Expect(step()).To(Succeed())
		}
	}

	It("marks inject-cloud-config Ignored=true for a container-image source", func() {
		d := deployer.NewDeployer(
			schema.Config{
				ISO:   schema.ISO{OverlayISO: "/tmp/overlay"},
				State: "/tmp/out",
			},
			schema.ReleaseArtifact{ContainerImage: "quay.io/kairos/rootfs:latest"},
			herd.EnableInit,
		)
		registerBuildISOSteps(d)

		injectCC := findEntry(d, constants.OpInjectCC)
		Expect(injectCC).NotTo(BeNil())
		Expect(injectCC.Ignored).To(BeTrue(),
			"herd disables StepInjectCC whenever ContainerImage is set, "+
				"so adding it to build-iso.go's step list is a no-op.")
	})

	It("marks inject-cloud-config Ignored=true for a dir: source as well", func() {
		d := deployer.NewDeployer(
			schema.Config{
				ISO:   schema.ISO{OverlayISO: "/tmp/overlay"},
				State: "/tmp/out",
			},
			schema.ReleaseArtifact{ContainerImage: "dir:/tmp/rootfs"},
			herd.EnableInit,
		)
		registerBuildISOSteps(d)

		injectCC := findEntry(d, constants.OpInjectCC)
		Expect(injectCC).NotTo(BeNil())
		Expect(injectCC.Ignored).To(BeTrue())
	})
})
```

Expected output:

```
Ran 2 of N Specs in ~0.001 seconds
SUCCESS! -- 2 Passed | 0 Failed
```

`Ignored=true` is not a proxy signal: `herd/ops_options.go:81-91` sets it when the `EnableIf` predicate is false, and `herd/dag.go:122` explicitly skips any Ignored entry's callback during `Run`.

So, 2 things to consider:

1. If `--overlay-iso` really is silently dropped on the subcommand path today, this PR does not fix it. The root cause has to be somewhere else, most likely in `ops.GenISO` / `BuildISOAction.ISORun` around `pkg/ops/iso.go:167-169` where the overlay is appended to `spec.Image` and applied into `isoDir` before `burnISO`. Worth reproducing the original repro on `main` and instrumenting from there.
2. The `pkg/ops/iso.go` refactor of `InjectISO` (running xorriso on overlay-only inputs) is a good defensive fix on its own, and its new test at `pkg/ops/inject_iso_test.go` passes. It just will not be reached through the build-iso subcommand until the herd guard is either removed or the subcommand starts constructing a deployer where `isoOption()` returns true.

Same check applied to `deployer.RegisterAll` (pipeline mode) with a container_image source produces `inject-cloud-config Ignored=true` too, so the PR body's claim that pipeline mode's `Deployer.Run` runs `StepInjectCC` also does not hold. Pipeline mode's overlay must be landing on the ISO via `GenISO`'s `spec.Image`, not `InjectISO`.

vipsharm:
Closing — the diagnosis behind this PR was wrong.

@jimmykarily was right in the review: `StepInjectCC` is `Ignored=true` on the `build-iso` subcommand path (and on the pipeline path too when a source is set), so wiring it into the step list is a no-op. The ISO code path was never the problem.

## Actual root cause

`urfave/cli` v2 follows Go stdlib `flag` semantics — **flag parsing stops at the first positional argument**. Our downstream Earthfile was invoking:

```
auroraboot build-iso dir:/build/image --overlay-iso /overlay --arch amd64
                     ^^^^^^^^^^^^^^^^ positional first
                                      ^^^^^^^^^^^^^^^^^^^ silently accepted as extra positional args and dropped
```

`--overlay-iso` (and `--arch`, and any subsequent flag) were being silently discarded. The ISO builds without them, `--overlay-iso` content never reaches `spec.Image`, and the produced ISO looks like it was built without any overlay. No `StepInjectCC` change needed — the value simply never made it into the schema.

Same silent-drop bug on `build-uki` for `--sb-key`, `--sb-cert`, `--tpm-pcr-private-key` — those all fail with `Required flags … not set` when the positional argument comes first.

## What we fixed

Downstream (CanvOS spectrocloud/CanvOS#749, spectrocloud/CanvOS#750): reordered every invocation to put `dir:...` LAST. AuroraBoot v0.26.2 subcommand mode then correctly honours `--overlay-iso`, produces a proper GPT-hybrid ISO with the CD-variant grub, no post-processing needed.

## The InjectISO refactor + test in this PR

Might still be useful on its own merits (`InjectISO` currently gates the whole xorriso call on cloud-config existence, so overlay-only pipeline-mode callers still see nothing land). But the primary framing of this PR is wrong so I'll withdraw it — happy to open a smaller PR that's just the InjectISO cleanup if maintainers think it's worth having.

## What I'd suggest as the real upstream fix

Reject unexpected positional args on `build-iso` and `build-uki` subcommands so silent flag drops fail loudly. This whole thread was misdiagnosed for about a week because the failure was a silent one. If a maintainer thinks that's welcome I'll file it as a fresh PR — no strong opinion on the shape (`ctx.NArg() > 1` check + error, or configure the cli app to reject unknown positional args globally).

Apologies for the noise on this one. Closing.

jimmykarily:
It's a real problem that it accepted the extra args and ignored them. It's so easy to get this wrong. Worth a fix.

PR #717, opened 2026-08-06
feat: AuroraBoot post-reboot reset lifecycle (kairos#4255)

wrkode:
## Summary

Implements the AuroraBoot server-side **post-reboot reset lifecycle** — the remaining piece of kairos-io/kairos#4255 after the agent side landed (kairos-agent #1307 selects the `statereset` boot entry and reboots) and boot-state reporting landed (#1317 + #655). Design agreed on the issue; @mudler-agent confirmed they're not working on it.

**The gap:** a `reset` reboots the node, so it can't complete synchronously — the command acks immediately, meaning "reset **scheduled**", not "reset **done**". AuroraBoot stored the reported `bootState` but never acted on it, so there was no confirmation the reset actually happened and no `lastReset`.

## Design — mirrors the existing `Deployment.EjectState` CAS

- `ManagedNode` gains **`ResetState`** (`"" | pending | in-progress | done | failed`), **`ResetRequestedAt`**, and **`LastReset`**.
- Issuing a `reset` command marks the node **`pending`** (best-effort; doesn't invalidate the queued command).
- On **re-register**, the lifecycle is resolved from the reported boot state. The phonehome agent calls `Register` on every boot, so a `Register` from a `pending`/`in-progress` node means it rebooted:
  - `autoreset` → **in-progress** (booted the statereset entry, mid-reset)
  - `active` → **done**, stamping `LastReset` (reset finished, node healthy)
  - `passive` / `recovery` → **failed** (the active image didn't survive)
- Transitions are compare-and-set (`... WHERE reset_state IN fromStates`) so concurrent re-registers resolve **exactly once** — same idiom as `ClaimForDelivery` / `CASEjectState`, correct on SQLite and PostgreSQL. `ResetState`/`LastReset` now surface in `GET /nodes` and `GET /nodes/:id`.

## Tests

Store (Ginkgo, file-backed SQLite): pending stamping, every transition, the `LastReset` stamp on done, the CAS no-op when the state isn't in `fromStates`, and exactly-one-winner under concurrent `AdvanceReset` (race-clean). Handlers: the command→pending hook (and that a non-reset command doesn't set it) plus every re-register resolution branch. Backward compatible — new columns default empty.

## Left for follow-up (noted on #4255)

A timeout sweep for a node that never comes back (`pending` → `failed` after a deadline), and the analogous `upgrade` lifecycle. Kept out to keep this focused on `reset`.

Closes the code side of kairos-io/kairos#4255.

### DIFF
--- a/docs/docs.go
+++ b/docs/docs.go
@@ -1968,6 +1968,10 @@ const docTemplate = `{
                 "lastHeartbeat": {
                     "type": "string"
                 },
+                "lastReset": {
+                    "description": "LastReset is when the most recent automatic reset completed successfully.",
+                    "type": "string"
+                },
                 "machineID": {
                     "type": "string"
                 },
@@ -1978,6 +1982,14 @@ const docTemplate = `{
                     }
                 },
                 "phase": {
+                    "type": "string"
+                },
+                "resetRequestedAt": {
+                    "description": "ResetRequestedAt is when the reset command was issued (ResetState set to\npending); nil when no reset has been requested.",
+                    "type": "string"
+                },
+                "resetState": {
+                    "description": "ResetState tracks the day-2 automatic-reset lifecycle across the reboot a\nreset command triggers (kairos-io/kairos#4255). A reset can't complete\nsynchronously — the agent selects the statereset boot entry and reboots — so\nthe command acks immediately and the real outcome is resolved here when the\nnode re-registers and reports its post-reboot boot state. One of:\n\"\" (none) | pending | in-progress | done | failed.",
                     "type": "string"
                 },
                 "updatedAt": {
--- a/docs/swagger.json
+++ b/docs/swagger.json
@@ -1961,6 +1961,10 @@
                 "lastHeartbeat": {
                     "type": "string"
                 },
+                "lastReset": {
+                    "description": "LastReset is when the most recent automatic reset completed successfully.",
+                    "type": "string"
+                },
                 "machineID": {
                     "type": "string"
                 },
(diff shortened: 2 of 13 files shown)


mudler-agent (changes requested):
The reset lifecycle is only initialized in `Create`. Both `CreateBulk` and `CreateForGroup` also create and deliver reset commands, but neither calls `SetResetPending`, so resets issued through `/api/v1/nodes/commands` or `/api/v1/groups/:id/commands` remain with an empty `ResetState` and the re-register resolver will ignore them. Please apply the pending-state hook to every command creation path and add reset coverage for the bulk and group endpoints.

mauromorales (approved):

wrkode:
Good catch, thanks @mudler-agent — fixed in 8bf4d45. `SetResetPending` is now applied on **all three** command-creation paths: `Create`, `CreateBulk` (`POST /api/v1/nodes/commands`), and `CreateForGroup` (`POST /api/v1/groups/:id/commands`), so a reset issued through the bulk or group endpoint is tracked the same as a single-node one and the re-register resolver picks it up. Added handler coverage for reset via bulk and group (plus a non-reset-bulk case that must leave `ResetState` empty).

mudler-agent (approved):
The bulk and group command paths now mark every selected node reset-pending after command creation, matching the single-node path, and the new tests cover both reset paths plus the non-reset case. The targeted store tests passed locally; the handler package could not redownload a blocked kernel.org module in this environment, while the PR unit-tests check is green. No remaining blocking findings.

PR #724, opened 2026-08-11
fix(ui): preserve Hadron filesystem layout

mudler-agent:
## Summary

- preserve the Hadron base’s `/usr/sbin` layout
- keep Docker/BuildKit copies working with older Hadron bases by replacing only a symlinked `/usr/local/lib/firmware` destination
- leave `/lib/firmware` and existing real firmware directories unchanged
- test generated middle content, caller ordering, and empty input

## Testing

- `npm test` (33 tests)
- `npm run build`
- `npm run lint` (0 errors, 3 existing warnings)

Related to kairos-io/kairos#4290. The current hadron and hadron-firmware images deliberately store firmware beneath `/usr/local`, so moving firmware to `/usr/lib` requires a coordinated upstream layout change and is not claimed by this PR.


mudler-agent:
Rebased onto current `main` at `79ad3f4` to refresh CI. Fresh local verification: all 48 UI tests pass, the production build passes, lint reports 0 errors with the same 3 existing warnings, and `git diff --check` passes. The focused Go package test could not start because this environment received HTTP 403 while downloading Go 1.26.6. CI has started on the rebased head and will be checked on the next cron run.

PR #726, opened 2026-08-11
feat(ui): make Hadron a peer flavor in the artifact builder

jimmykarily:
Hadron used to sit at position 7 in the template grid and hijacked Step 0 with a bespoke compose sub-wizard whose primary CTA was the only way forward, so the Standard variant with Kubernetes on the image was reachable from Ubuntu but not from Hadron even though the variant filter was never per-flavor to begin with.

- Move Hadron to the first tile with the same prefill shape as Ubuntu (Renovate-tracked Hadron base, variant core, model generic, ISO output).
- Auto-advance to Configure on click, matching every other flavor.
- Repackage the firmware, software-layers, extra-Dockerfile, and version pickers as an optional collapsible "Advanced" card at the tail of Configure. Composition triggers on Next when any field is populated; otherwise the plain image path runs.
- Sync form.baseImage at the composer's version setters instead of via a derived effect to satisfy react-hooks/set-state-in-effect.
- Preserve kubernetesDistro, kubernetesVersion, and kubernetesEnabled when cloning a Hadron artifact (previously dropped silently in the Hadron clone branch) and land Hadron clones on Configure with the expander pre-opened.

Adds artifactBuilder.hadron.test.tsx covering the Hadron-first ordering, auto-advance with Standard reachable, the Kubernetes card reveal, and the clone Kubernetes preservation.

### DIFF
--- a/internal/builder/auroraboot/builder.go
+++ b/internal/builder/auroraboot/builder.go
@@ -198,6 +198,7 @@ func (b *Builder) Build(ctx context.Context, opts builder.BuildOptions) (*builde
 
 	// Persist artifact record in DB if store is available.
 	if b.store != nil {
+		kubernetesEnabled := opts.Provisioning.KubernetesEnabled
 		rec := &store.ArtifactRecord{
 			ID:                      id,
 			Name:                    opts.Name,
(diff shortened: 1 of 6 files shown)


jimmykarily:
image

and expanded:

image

mauromorales (approved):

PR #729, opened 2026-08-14
fix(security): constant-time comparison for admin/node bearer tokens

wrkode:
Routes the admin-password / node-API-key checks in `AdminMiddleware`, `AgentOrAdminMiddleware`, and `DownloadMiddleware` through a `secureCompare` helper backed by `crypto/subtle.ConstantTimeCompare`, instead of plain string `==` (which returns on the first differing byte and leaks, via response timing, how many leading bytes of the secret a caller guessed).

Behaviour is unchanged — a correct token still authenticates, a wrong or empty one is still rejected — so it's transparent to every API client (nodes, UI, the fleet/CAPI integration); only the timing side-channel is closed. Adds an internal unit test for the helper (equal / unequal / unequal-length / empty).

Refs: kairos-io/kairos#4117 (fleet-server hardening, 'smaller but real').

_Note: the GCE/VHD raw-disk race this PR originally also carried was reworked into the copy-based fix in #730 (which covers GCE, VHD and MAAS), so this PR is now the constant-time change alone._

### DIFF
--- a/pkg/auth/middleware.go
+++ b/pkg/auth/middleware.go
@@ -2,6 +2,7 @@ package auth
 
 import (
 	"bytes"
+	"crypto/subtle"
 	"encoding/json"
 	"io"
 	"net/http"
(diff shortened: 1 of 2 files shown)


jimmykarily (approved):

mudler-agent:
Stale triage: this fix still applies and is not superseded. Current `main` still compares the admin bearer token with plain `==` in `AdminMiddleware`, `AgentOrAdminMiddleware`, and `DownloadMiddleware`, while this PR routes those checks through `crypto/subtle.ConstantTimeCompare`. The PR is mergeable, was approved by @jimmykarily on August 18, and its original required checks passed. This should remain open for merge rather than be closed as obsolete.

mauromorales (approved):

PR #730, opened 2026-08-17
fix(build): convert raw disks on a private copy to avoid the shared-file race

wrkode:
Fixes the shared-raw-disk race across **all three** cloud-image conversions (GCE, VHD, MAAS), by working on a private copy — per @jimmykarily's review suggestion, which is both cleaner and more correct than the ordering approach this PR originally used.

## The race

GCE, VHD and MAAS each globbed the single `kairos-*.raw` and operated on it directly, so a build requesting more than one ran them in parallel over the same file. They touch it incompatibly:

| Conversion | Effect on the raw |
|---|---|
| GCE | **truncates** in place (pad to GB) |
| VHD | **renames** it away (+ VHD footer) |
| MAAS | **reads** it (gzip) |

Racing them corrupts the image, or a reader loses to a rename (`no such file`). And because GCE/VHD mutate the raw **in place**, they also corrupt the raw itself when the plain `.raw`/EFI disk is a requested output — which pure step-ordering does not fix.

## The fix (work on a copy, not serialize)

`convertRawOnCopy` copies the raw into a throwaway subdirectory, runs the converter (`Raw2Gce`/`Raw2Azure`/`Raw2MAAS`) against the **copy**, and moves the produced artifact back next to the raw under its conventional `<raw>.<ext>` name. The original raw is never mutated, so:

- the conversions are safe to run **concurrently** (no serialization, faster than ordering);
- the raw itself stays a valid deliverable;
- the DAG-ordering complexity is gone (this replaced the earlier `ConditionalOption` approach).

The `Raw2Gce`/`Raw2Azure`/`Raw2MAAS` primitives are unchanged — only the `ConvertRawDiskTo*` wrappers now route through the copy helper. The GCE/VHD wrappers already created (and never used) a `gendisk` temp dir; this is the finished version of that intent.

## Tests

Ginkgo (`pkg/ops`, matching the suite there): `convertRawOnCopy` runs the converter on a copy and leaves the original raw **byte-for-byte**; propagates converter errors without touching the raw; and — the direct proof — keeps the shared raw intact when two destructive converters (one truncating, one renaming) run **concurrently**. `ConvertRawDiskToMAAS` produces `<raw>.gz` with the raw preserved. `go build`/`vet` and the `pkg/ops` + `deployer` suites green.

Refs: kairos-io/kairos#4117.

### DIFF
--- a/pkg/ops/convert_copy_test.go
+++ b/pkg/ops/convert_copy_test.go
@@ -1,0 +1,112 @@
+package ops
+
+import (
+	"context"
+	"errors"
+	"os"
+	"path/filepath"
+	"sync"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("raw-disk conversion on a private copy", func() {
+	const rawContent = "original-raw-bytes-that-must-survive-conversion"
+	var dir, raw string
+
+	BeforeEach(func() {
+		dir = GinkgoT().TempDir()
+		raw = filepath.Join(dir, "kairos-test.raw")
+		Expect(os.WriteFile(raw, []byte(rawContent), 0o644)).To(Succeed())
+	})
+
+	Describe("convertRawOnCopy", func() {
+		It("runs the converter on a copy and leaves the original raw untouched", func() {
+			// A converter that destroys its source (the way Raw2Gce truncates and
+			// Raw2Azure renames the raw in place). If it were handed the original
+			// instead of a copy, the raw-survives assertion below would fail.
+			convert := func(source string) (string, error) {
+				Expect(source).NotTo(Equal(raw), "converter must receive a copy, not the original raw")
+				Expect(os.Remove(source)).To(Succeed())
+				out := source + ".converted"
+				return out, os.WriteFile(out, []byte("output"), 0o644)
+			}
+
+			dst, err := convertRawOnCopy(dir, convert)
+			Expect(err).NotTo(HaveOccurred())
+
+			// The output was moved next to the raw under <raw>.<ext>.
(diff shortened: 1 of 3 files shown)


jimmykarily:
Can't the steps act on a copy of the file instead of mutating for all the other steps? (unless I didn't get it right). Also, please switch to ginkgo-style tests for consistency with the rest of the test suites in all projects.

pkg/ops/convert_copy_test.go
```diff
@@ -0,0 +1,112 @@
...
+			Expect(err).NotTo(HaveOccurred())
+
+			// The output was moved next to the raw under <raw>.<ext>.
+			Expect(dst).To(Equal(filepath.Join(dir, "kairos-test.raw.converted")))
+			Expect(dst).To(BeAnExistingFile())
+
+			// The original raw survived, byte-for-byte.
+			got, err := os.ReadFile(raw)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(string(got)).To(Equal(rawContent))
```
jimmykarily, line 46: I would be good if we also tested that the "converted" file is not similar to `rawContent`. This is to ensure that the original is maintained because we use a copy and not because the whole call was a no-op.

jimmykarily (approved):

mudler-agent:
Stale triage: this is still relevant and is not superseded by current `main`. The current `ConvertRawDiskToVHD` and `ConvertRawDiskToGCE` paths still pass the shared `glob[0]` directly to `Raw2Azure` and `Raw2Gce`; there is no equivalent private-copy helper on `main`.

The requested copy-based approach and Ginkgo tests are present here, Jimmy approved the current head, and all functional CI checks passed. Only patch coverage is red. This should be refreshed for merge rather than closed as obsolete.

PR #739, opened 2026-08-20
fix(iso): materialize atomic-writer overlays

### LINKED ISSUE #4324: auroraboot build-iso: --overlay-iso fails with exit status 5 when overlay dir is a Kubernetes Secret/ConfigMap volume mount (symlink structure)
### Description

`auroraboot build-iso --overlay-iso <dir> dir:<rootfs>` fails with `Failed creating ISO image: exit status 5` when the overlay directory is a Kubernetes Secret or ConfigMap volume mount (i.e. contains the kubelet "atomic writer" symlink structure), even though the same overlay content succeeds when copied as plain files/directories.

This affects any consumer that wires a Kubernetes-mounted Secret/ConfigMap directly into `--overlay-iso` (e.g. `kairos-operator`'s `OSArtifact.spec.overlayISOVolume`), which is the natural way to supply overlay files (like a custom `grub.cfg`) in a Kubernetes-native build pipeline.

### Root cause (confirmed via local reproduction)

Kubernetes mounts Secret/ConfigMap volumes using an "atomic writer" layout: each top-level path component is a symlink into a hidden, timestamped directory, e.g.:

```
overlay-iso/
├── ..data -> ..2026_08_17_20_28_56.3347250082
├── ..2026_08_17_20_28_56.3347250082/
│   └── boot/grub2/grub.cfg
└── boot -> ..data/boot          # <-- top-level entry is a SYMLINK, not a real dir
```

When `auroraboot` copies `--overlay-iso` onto the already-populated ISO root tree (which by that point contains a real `boot/` directory created by the earlier "Creating EFI image" step), the overlay's top-level `boot` symlink collides with the destination's real `boot/` directory.

Reproducing the merge manually with plain `rsync` inside the `auroraboot` container shows the collision explicitly:

```
$ rsync -av /overlay-iso/ /iso-base/
could not make way for new symlink: boot
cannot delete non-empty directory: boot
rsync error: some files/attrs were not transferred (see previous errors) (code 23)
```

`auroraboot`'s own overlay-copy step (`[1] Copying /overlay-iso source to .../iso`) does **not** surface this collision — it logs `Finished syncing` cleanly regardless — but the ISO root tree's `/boot` ends up broken/shadowed by the time `xorriso` runs, and `xorriso` then fails while building the boot catalog/El Torito image with `exit status 5`.

### Reproduction

Two directories, identical content, only difference is whether `boot` is a plain directory or a Kubernetes-style symlink:

**Control — plain overlay (succeeds):**
```
overlay-plain/
└── boot/grub2/grub.cfg      # real directory
```
```
docker run --rm --privileged \
  -v "$PWD/rootfs:/rootfs:ro" \
  -v "$PWD/overlay-plain:/overlay-iso" \
  -v "$PWD/out-plain:/artifacts" \
  quay.io/kairos/auroraboot:v0.24.0 \
  --debug build-iso --override-name test-plain --date=false --output /artifacts --arch amd64 \
  --overlay-iso /overlay-iso dir:/rootfs
```
→ exits 0, produces a valid `test-plain.iso`.

**Test — Kubernetes-style symlinked overlay (fails):**
```
overlay-symlinked/
├── ..2026_08_17_20_28_56.3347250082/boot/grub2/grub.cfg
├── ..data -> ..2026_08_17_20_28_56.3347250082
└── boot -> ..data/boot
```
```
docker run --rm --privileged \
  -v "$PWD/rootfs:/rootfs:ro" \
  -v "$PWD/overlay-symlinked:/overlay-iso" \
  -v "$PWD/out-symlinked:/artifacts" \
  quay.io/kairos/auroraboot:v0.24.0 \
  --debug build-iso --override-name test-symlinked --date=false --output /artifacts --arch amd64 \
  --overlay-iso /overlay-iso dir:/rootfs
```
→ exits 1 with:
```
INF [1] Creating ISO image...
ERR [1] Failed creating ISO image: exit status 5
ERR Failed generating iso 'test-symlinked' from '...'. Error: 1 error occurred:
	* exit status 5
```

Both tests use `quay.io/kairos/auroraboot:v0.24.0`, run via `docker` (verified with `colima` as the local daemon on macOS). The symlink structure in `overlay-symlinked/` was captured verbatim from `find`/`ls -la` inside a real pod with the overlay Secret mounted at the same path — this is not a synthetic/contrived layout, it's exactly what `kubelet` produces for every Secret/ConfigMap volume mount.

### Expected behavior

`--overlay-iso` should support directories that are (or contain) Kubernetes Secret/ConfigMap volume mounts, since that's the standard way to supply files to a container in Kubernetes-native pipelines like `kairos-operator`'s `OSArtifact`. At minimum, the overlay-copy step should either:
- dereference symlinks when copying (`cp -L` / `rsync --copy-links` semantics) instead of trying to recreate them at the destination, or
- merge into existing destination directories instead of attempting a full directory→symlink replacement, or
- fail loudly at the copy step instead of silently reporting `Finished syncing` and failing later, opaquely, at the `xorriso` step.

### Environment

- `auroraboot` image: `quay.io/kairos/auroraboot:v0.24.0`
- Triggered via `kairos-operator`'s `OSArtifact.spec.volumes` / `overlayISOVolume` (Secret-backed), but reproduced standalone with plain `docker run` (no Kubernetes involved) as shown above.

mudler-agent:
## What changed

- Materialize Kubernetes atomic-writer overlay links before ISO copies.
- Apply the staging path to generated and downloaded ISO workflows.
- Preserve ordinary links, dot-prefixed entries, and file modes.
- Reject links that resolve outside the overlay root.

## Why

Secret and ConfigMap mounts expose logical files through `..data` symlinks. The existing archive copy preserved those links, so a path such as `boot -> ..data/boot` collided with the ISO root's existing `boot/` directory.

## Verification

- Focused standard-library regression suite: PASS, 5 tests
- `git diff --check`: PASS
- Full `go test ./pkg/ops/...`: blocked before compilation because the environment returns HTTP 403 while downloading Go 1.26.6 from `proxy.golang.org`

Closes kairos-io/kairos#4324

### DIFF
--- a/pkg/ops/iso.go
+++ b/pkg/ops/iso.go
@@ -165,7 +165,12 @@ func GenISO(srcFunc, dstFunc valueGetOnCall, i schema.ISO) func(ctx context.Cont
 			spec.RootFS = append(spec.RootFS, imagetypes.NewDirSrc(i.OverlayRootfs))
 		}
 		if i.OverlayISO != "" {
-			spec.Image = append(spec.Image, imagetypes.NewDirSrc(i.OverlayISO))
+			materializedOverlay, cleanup, err := materializeISOOverlay(i.OverlayISO)
+			if err != nil {
+				return err
+			}
+			defer cleanup()
+			spec.Image = append(spec.Image, imagetypes.NewDirSrc(materializedOverlay))
 		}
 
 		buildISO := NewBuildISOAction(cfg, spec)
@@ -190,7 +195,12 @@ func InjectISO(dstFunc, isoFunc valueGetOnCall, i schema.ISO) func(ctx context.C
 
 		if i.OverlayISO != "" {
 			internal.Log.Logger.Info().Msgf("Adding overlay data in '%s' to '%s'", i.OverlayISO, isoFile)
-			err = copy.Copy(i.OverlayISO, tmp)
+			materializedOverlay, cleanup, materializeErr := materializeISOOverlay(i.OverlayISO)
+			if materializeErr != nil {
+				return materializeErr
+			}
+			defer cleanup()
+			err = copy.Copy(materializedOverlay, tmp)
 			if err != nil {
 				return err
 			}
--- a/pkg/ops/iso_overlay.go
+++ b/pkg/ops/iso_overlay.go
@@ -1,0 +1,175 @@
+package ops
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+func materializeISOOverlay(source string) (string, func(), error) {
+	root, err := filepath.EvalSymlinks(source)
+	if err != nil {
+		return "", nil, fmt.Errorf("resolve ISO overlay root: %w", err)
+	}
+	root, err = filepath.Abs(root)
+	if err != nil {
+		return "", nil, fmt.Errorf("make ISO overlay root absolute: %w", err)
+	}
+
+	target, err := os.MkdirTemp("", "auroraboot-iso-overlay-")
+	if err != nil {
+		return "", nil, fmt.Errorf("create ISO overlay staging directory: %w", err)
+	}
+	cleanup := func() {
+		_ = filepath.WalkDir(target, func(path string, entry os.DirEntry, err error) error {
+			if err == nil && entry.IsDir() {
+				_ = os.Chmod(path, 0o700)
+			}
+			return nil
+		})
+		_ = os.RemoveAll(target)
+	}
+
+	activeDirs := map[string]bool{}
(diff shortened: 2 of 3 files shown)


mudler-agent:
Cron CI triage: `unit-tests` failed in the Run tests step and `build-linux-riscv64` failed in Build binary. All image, e2e, boot, Helm, security, and DCO checks passed. GitHub returns 403 for both job-log downloads, and annotations contain only exit code 1, so there is not enough evidence for a code change. I attempted to rerun both failed jobs, but `mudler-agent` lacks repository Actions rerun permission. A maintainer rerun of runs 32404743476 and 32404743567 is needed.

PR #742, opened 2026-08-20
Fix ISO Overlay when the files are symlinks.

ebourgeois:
Fix ISO Overlay when the files are symlinks.

Fixes: #4324

### DIFF
--- a/pkg/ops/inject_iso_test.go
+++ b/pkg/ops/inject_iso_test.go
@@ -58,6 +58,77 @@ var _ = Describe("InjectISO", Label("iso"), func() {
 		Expect(string(lsOut)).To(ContainSubstring("config.yaml"))
 	})
 
+	It("injects a Kubernetes-mounted (symlinked) overlay as plain files", func() {
+		// Regression test for https://github.com/kairos-io/kairos/issues/4324:
+		// kairos-operator's OSArtifact.spec.overlayISOVolume mounts a
+		// Secret/ConfigMap directly as the overlay, which arrives in kubelet's
+		// "atomic writer" layout — every top-level entry a symlink into a
+		// hidden ..data directory. Injected verbatim, those symlinks end up in
+		// the ISO root (and can shadow real dirs like boot/).
+		if _, err := exec.LookPath("xorriso"); err != nil {
+			Skip("xorriso not installed")
+		}
+
+		dst := GinkgoT().TempDir()
+
+		// The overlay in the exact kubelet layout:
+		//   overlay/..data -> ..2026_08_17_20_28_56.3347250082
+		//   overlay/..2026_08_17_20_28_56.3347250082/boot/grub2/grub.cfg
+		//   overlay/boot -> ..data/boot
+		overlay := GinkgoT().TempDir()
+		timestamped := filepath.Join(overlay, "..2026_08_17_20_28_56.3347250082")
+		Expect(os.MkdirAll(filepath.Join(timestamped, "boot", "grub2"), 0o755)).To(Succeed())
+		Expect(os.WriteFile(filepath.Join(timestamped, "boot", "grub2", "grub.cfg"),
+			[]byte("set timeout=5\n"), 0o644)).To(Succeed())
+		Expect(os.Symlink(filepath.Base(timestamped), filepath.Join(overlay, "..data"))).To(Succeed())
+		Expect(os.Symlink(filepath.Join("..data", "boot"), filepath.Join(overlay, "boot"))).To(Succeed())
+
+		// Build a minimal source ISO so xorriso has a real -indev/-outdev.
+		payload := GinkgoT().TempDir()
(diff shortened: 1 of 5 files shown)


mauro-agent (commented):
🤖 **Automated review by `mauro-agent`** — an AI agent operated by `@mauromorales`. This is not Mauro's personal review; he has not read this PR unless he says so himself. If you want a human, feel free to ping him directly.

Static review only; I did not build this or run the tests.

**Summary**: Fixes kairos-io/kairos#4324 by dereferencing symlinks and dropping kubelet's atomic-writer internals (`..data`, `..<timestamp>`) when an `--overlay-iso` directory is merged onto the ISO root — in `GenISO`, `InjectISO`, and the UKI ISO path. The new `pkg/utils/overlay.go` centralizes the logic and all three call sites use it consistently. Checked the cleanup ordering at each call site: the temp overlay directory is only removed after the synchronous build/copy call that consumes it, so the `defer os.RemoveAll(...)` placement is correct in all three.

**Question, not a blocker**: `OverlayCopyOptions`'s `OnSymlink` always deep-dereferences every symlink in the tree (`copy.Deep`), not only the kubelet `..data`-style top-level ones. A symlink anywhere in the overlay that resolves outside the overlay root — an absolute path, or a `../..` escape — would have its target's content copied verbatim into the built image. Is that an accepted trade-off given the overlay source is already a trusted mount (operator-provided Secret/ConfigMap, or a local directory the invoking user controls), or would constraining dereferenced targets to stay under the overlay root be worth adding?

**FYI, unrelated to this PR's diff**: `InjectISO`'s `xorriso -map` step, which actually merges the overlay onto the built ISO, only runs when `dst/config.yaml` exists and is non-empty — that gate predates this change. An `--overlay-iso` supplied without a cloud-config looks like it is silently dropped in that code path today. This PR doesn't touch that gate and the new regression test only exercises the config.yaml-present case, so flagging for awareness rather than as something this PR needs to fix.

Test coverage looks thorough: the exact kubelet layout (directory symlink, file symlink, mixed content, nested symlink, dangling-symlink failure mode) plus an end-to-end ISO-build regression test for #4324 itself.

ebourgeois:
> **FYI, unrelated to this PR's diff**: `InjectISO`'s `xorriso -map` step, which actually merges the overlay onto the built ISO, only runs when `dst/config.yaml` exists and is non-empty — that gate predates this change. An `--overlay-iso` supplied without a cloud-config looks like it is silently dropped in that code path today. This PR doesn't touch that gate and the new regression test only exercises the config.yaml-present case, so flagging for awareness rather than as something this PR needs to fix.

@mauromorales, I agree on the complete dereferencing and I have pushed a fix for that too. Regarding the informational comment only, i.e. the gate, is this expected behaviour? Else, I can fix that too, as  it's slightly worse than described. In `pkg/ops/iso.go`, InjectISO:

   • iso.go:199-207, the overlay is staged into tmp whenever i.OverlayISO != "". Good.
   • iso.go:210-233, the xorriso -indev … -map tmp / that merges tmp into the ISO only runs inside the config.yaml exists && non-empty branch.

   So with `--overlay-iso` and no cloud-config, the overlay is copied into tmp and then discarded by the deferred RemoveAll without ever reaching the ISO. Two extra details worth adding to
   the note:

   1. It's silent twice. The "No cloud config file found … skipping injection" warn at iso.go:231 only fires when config.yaml exists but is empty. When the file is simply absent — the
      normal overlay-only case — the outer os.Stat fails and nothing is logged at all. An overlay-only build completes green with zero signal.
   2. The natural follow-up fix is small: gate the xorriso run on "tmp has content" (overlay staged or config.yaml copied) rather than on config.yaml presence, e.g. track a staged bool
      set by both branches. The regression-test gap the user mentioned is real: all three tests in pkg/ops/inject_iso_test.go (:27, :94-97, :134) write a config.yaml before invoking
      InjectISO, so the overlay-only path has no coverage.

   Leaving it untouched per your flag, it predates fix-iso-overlay-copy and doesn't belong in that PR's diff. If you want it tracked outside the PR, it'd make a good upstream
   kairos-io/auroraboot issue (the symlink-overlay work already references kairos-io/kairos#4324, so the maintainers are clearly receptive to this area getting attention); say the word
   and I'll draft one.

mauro-agent:
🤖 **Follow-up from `mauro-agent`**, the same automated reviewer as above.

Checked `checkSymlinkContained` against the concern: it resolves the symlink target fully (through further symlinks too, via `EvalSymlinks`), compares against the resolved overlay root, and rejects anything that lands outside it, an absolute-path escape, a `..` escape, or a symlink to its own ancestor (that last one also closes an infinite-recursion path under `copy.Deep` I hadn't even asked about). Four dedicated tests cover it: the two escape cases, the self-ancestor case, and a check that a legitimate absolute symlink *inside* the root still works.

That resolves my open question. Nothing else outstanding from this review.

mauromorales (approved):

PR #762, opened 2026-08-31
fix(serve): harden the netboot artifact server — no directory listing + unauthenticated-serve warning

wrkode:
Hardens the unauthenticated netboot/artifact HTTP servers, addressing the "serve" line item of the fleet-server hardening epic (kairos-io/kairos#4117). Two related changes:

## 1. No browsable directory listing

`ServeArtifacts` used a bare `http.FileServer(http.Dir(dir))`, which renders a browsable index for any directory request — a `GET /` on the netboot server leaked every filename in the artifact directory to anyone who could reach it on `0.0.0.0`. A `noListingFS` wrapper now returns `os.ErrNotExist` for directories, so files are still served by their exact path but the root (and any subdirectory) 404s instead of listing.

## 2. Warn that the serve is unauthenticated (serve-auth)

These servers exist so a **PXE/HTTP-booting machine** can fetch its kernel, initrd, squashfs and the cloud-config baked into the ISO — over plain HTTP. A booting machine has **no way to present a credential**, so the servers *cannot* be gated behind a bearer token without breaking netboot. The real security boundary is the network, not authentication.

Rather than leave that implicit, both serve paths (`ServeArtifacts` and the UKI PXE ISO server) now log a prominent **warning at startup**: the server is unauthenticated, the served cloud-config may carry a registration token / SSH keys / passwords, and it must run on a **trusted, isolated provisioning network** — flagging when it binds all interfaces. A README note documents the boundary and points operators at `--set listen_addr=<host>:8080` to scope the artifact server to a specific interface.

The shared warning helper lives in `internal` (imported by both serve paths); a table test covers the all-interfaces address classification. Existing `ServeArtifacts` Ginkgo tests assert a known file is still served while the root 404s and does not leak filenames.

> **Why not real auth?** A token-in-URL scheme (isoserve-style opaque path baked into the pixiecore boot script) is the only PXE-compatible way to gate these servers, and it's a larger cross-cutting change through the `kairos-io/netboot`/pixiecore integration. That's tracked as a possible follow-up; this PR closes the item at the "documented boundary + no listing" level.

Refs: kairos-io/kairos#4117

### DIFF
--- a/internal/serve_warn.go
+++ b/internal/serve_warn.go
@@ -1,0 +1,22 @@
+package internal
+
+// WarnUnauthenticatedServe logs a prominent warning that a netboot/artifact HTTP
+// server serves its files without authentication. These servers exist so a
+// PXE/HTTP-booting machine can fetch its kernel, initrd, squashfs and the
+// cloud-config baked into the image over plain HTTP, and a booting machine has no
+// way to present a credential — so the server cannot be gated behind a token
+// without breaking netboot. Anyone who can reach addr can therefore download
+// whatever is served, including a cloud-config that may carry a registration
+// token, SSH keys or passwords. Run it only on a trusted, isolated provisioning
+// network.
+//
+// The lack of authentication is the point being flagged; it is independent of
+// which interfaces the server binds. A server can bind all interfaces and still
+// require auth, or bind one interface and require none — so this warning speaks
+// only to the missing authentication, not to the listen address.
+func WarnUnauthenticatedServe(addr, name string) {
+	Log.Logger.Warn().
+		Str("address", addr).
+		Str("server", name).
+		Msg("Serving over UNAUTHENTICATED HTTP: booting machines cannot present a credential, so anyone who can reach this address can download the served files, which may include a cloud-config with secrets. Run this only on a trusted, isolated provisioning network — do not expose it to untrusted networks or the internet.")
+}
--- a/README.md
+++ b/README.md
@@ -131,6 +131,8 @@ docker run --rm -ti --net host quay.io/kairos/auroraboot \
 
 This downloads the needed artifacts, bakes your cloud-config into a custom ISO, and serves it over the network.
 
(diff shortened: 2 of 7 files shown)


internal/serve_warn.go
```diff
@@ -0,0 +1,39 @@
...
+// gated behind a token without breaking netboot. The security boundary is
+// therefore the network: anyone who can reach addr can download whatever is
+// served, including a cloud-config that may carry a registration token, SSH keys
+// or passwords. Run it only on a trusted, isolated provisioning network.
+func WarnUnauthenticatedServe(addr, name string) {
+	w := Log.Logger.Warn().Str("address", addr).Str("server", name)
+	if bindsAllInterfaces(addr) {
+		w = w.Bool("all_interfaces", true)
+	}
+	w.Msg("Serving over UNAUTHENTICATED HTTP: booting machines cannot present a credential, so anyone who can reach this address can download the served files, which may include a cloud-config with secrets. Run this only on a trusted, isolated provisioning network — do not expose it to untrusted networks or the internet.")
```
jimmykarily: Binding all interfaces doesn't necessarily mean the server doesn't require authentication. Ours doesn't require authentication but that's a different issue to just listening on all interfaces, no?

wrkode: Good catch — you're right, they're orthogonal. I've dropped the `bindsAllInterfaces`/`all_interfaces` logic entirely so the warning no longer conflates the two. It now speaks only to the missing authentication (the actual, always-true condition here, since a PXE-booting client can't present a credential), independent of which interfaces the server binds. The helper is down to a single `WarnUnauthenticatedServe` call, now covered by a buffer-logger test. (Force-pushed as 5c6f2c6.)

internal/serve_warn.go
```diff
@@ -0,0 +1,22 @@
...
+// way to present a credential — so the server cannot be gated behind a token
+// without breaking netboot. Anyone who can reach addr can therefore download
+// whatever is served, including a cloud-config that may carry a registration
+// token, SSH keys or passwords. Run it only on a trusted, isolated provisioning
+// network.
+//
+// The lack of authentication is the point being flagged; it is independent of
+// which interfaces the server binds. A server can bind all interfaces and still
+// require auth, or bind one interface and require none — so this warning speaks
+// only to the missing authentication, not to the listen address.
```
jimmykarily, line 16:
```suggestion
```

jimmykarily (approved):

PR #763, opened 2026-08-31
fix(config): surface --set parse errors instead of silently dropping them

wrkode:
Closes the "ignored `yaml.Unmarshal` errors in `internal/config`" item from the fleet-server hardening backlog (kairos-io/kairos#4117).

## The bug

`ReadConfig` applies `--set key=value` pairs by rendering them to YAML and unmarshalling into the `Config`, `ReleaseArtifact`, and template-value targets — then **discarded all three `yaml.Unmarshal` errors**. A `--set` value that doesn't fit its field's type (e.g. `disk.gce=notabool` for a boolean) was silently dropped: the field kept its zero value and the operator got no indication the override was ignored.

## The fix

Return those errors. This is safe for valid input — verified empirically, since the package had **no tests**:
- Well-typed `--set` values (`disk.gce=true`, `disk.size=2000`, …) type-coerce and apply cleanly, so they don't error.
- `Config` and `ReleaseArtifact` share **no field names**, so a `--set` meant for one is just an unknown key to the other (ignored, not an error). Only a genuine type mismatch on a field that target owns now fails.
- The e2e suite's `--set` usage is all well-typed, so it won't regress.

## Tests

First tests for the package (Ginkgo): well-typed values apply to config + release without error; a malformed value now surfaces an error (`ContainSubstring("--set")`); an empty option set is accepted.

Refs: kairos-io/kairos#4117.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -116,9 +116,21 @@ func ReadConfig(fileConfig, cloudConfig string, options []string) (*schema.Confi
 		return c, r, err
 	}
 
-	yaml.Unmarshal(y, c)
-	yaml.Unmarshal(y, r)
-	yaml.Unmarshal(y, &templateValues)
+	// Apply the --set values (as YAML) onto each target, surfacing type
+	// mismatches instead of dropping them. One loop rather than three copies so
+	// the single error path is exercised by any target's failure.
+	for _, target := range []struct {
+		dst   any
+		label string
+	}{
+		{c, "config"},
+		{r, "release artifact"},
+		{&templateValues, "template values"},
+	} {
+		if err := yaml.Unmarshal(y, target.dst); err != nil {
+			return c, r, fmt.Errorf("applying --set values to %s: %w", target.label, err)
+		}
+	}
 
 	if cloudConfig != "" {
 		var err error
--- a/internal/config/read_config_test.go
+++ b/internal/config/read_config_test.go
@@ -1,0 +1,52 @@
+package config_test
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/kairos-io/AuroraBoot/internal/config"
+)
+
+var _ = Describe("ReadConfig --set handling", func() {
+	It("applies well-typed --set values to the config and release without error", func() {
+		c, r, err := config.ReadConfig("", "", []string{
+			"arch=arm64",
+			"disable_iso=true",
+			"disk.gce=true",
+			"disk.size=2000",
+			"container_image=oci://example/img:tag",
+		})
+		Expect(err).NotTo(HaveOccurred())
+		Expect(c.Arch).To(Equal("arm64"))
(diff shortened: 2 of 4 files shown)


internal/config/config.go
```diff
@@ -116,9 +116,22 @@ func ReadConfig(fileConfig, cloudConfig string, options []string) (*schema.Confi
...
-	yaml.Unmarshal(y, c)
-	yaml.Unmarshal(y, r)
-	yaml.Unmarshal(y, &templateValues)
+	// Apply the --set values (as YAML) onto each target. Surface unmarshal
+	// errors instead of dropping them: a malformed value (e.g. a non-boolean for
+	// a boolean field) used to be silently ignored, leaving the field at its
+	// zero value and the operator none the wiser. Config and ReleaseArtifact
+	// share no field names, so a --set meant for one is simply an unknown key to
+	// the other (ignored, no error) — only a genuine type mismatch on a field
+	// that target owns produces an error.
```
jimmykarily:
```suggestion
	// Apply the --set values (as YAML) onto each target.
```

It was a mistake to ignore the errors. I don't think we need to explain why errors should be handled.

internal/config/read_config_test.go
```diff
@@ -0,0 +1,42 @@
...
+		// field kept its zero value, no error reported). It must now surface, so a
+		// typo in --set is not swallowed.
+		_, _, err := config.ReadConfig("", "", []string{"disable_iso=notabool"})
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(ContainSubstring("--set"))
+	})
+
+	It("accepts an empty option set", func() {
+		c, r, err := config.ReadConfig("", "", nil)
+		Expect(err).NotTo(HaveOccurred())
```
jimmykarily, line 48: Better add a `MatchError` expectation to make sure the error is the expected one (and not something irrelevant).

jimmykarily (approved):

internal/config/config.go
wrkode: Done — trimmed to your one-liner. (85b9a18)

internal/config/read_config_test.go
wrkode: Done — now `Expect(err).To(MatchError(ContainSubstring("applying --set values to config")))`, so it pins the expected wrapped error (from the config target specifically) rather than accepting any error. (85b9a18)

wrkode:
Heads up @jimmykarily — the failing `codecov/patch` was flagging the third unmarshal's error branch (`template values`), which is effectively unreachable: unmarshaling the already-validated `--set` YAML into a `map[string]interface{}` can't fail the way it can for the two typed structs. Rather than leave a permanently-uncovered defensive line, I consolidated the three identical checks into one loop over the targets, so the single error path is exercised (and fully covered) by any target's failure — all errors are still surfaced, same behaviour. Also added a test that a `--set` mismatching a release-artifact field surfaces from that target specifically (covers the second branch). Re-request review at your convenience.

PR #526, opened 2026-05-26
Feat/extensions design

mudler:
giphy

Depends on https://github.com/kairos-io/kairos-agent/pull/1226

PR #771, opened 2026-09-03
feat(extensions): rebase the sysext/confext work from #526 onto main

ci-robbot:
Re-applies the extensions work from #526 onto current `main`, as @mudler asked on https://github.com/kairos-io/AuroraBoot/pull/526#issuecomment-5518908558.

## Why this is 55 files and not 303

A straight rebase of `feat/extensions-design` would have been the wrong tool. That branch is 139 ahead / 181 behind, and a single three-way merge against `main` conflicts in 55 files across 250 hunks. Most of that is not the extensions feature fighting `main`, it is the branch's *own already-merged work* fighting `main`: of the 190 files the branch adds, **161 already exist on `main`**, landed over the last three months through other PRs (the operator builder, redfish vendor quirks, `internal/secrets`, the settings store, BMC encryption, deployment eject, and so on).

So instead of replaying 139 commits and re-resolving conflicts against work that is already upstream, this PR extracts the 29 files that are genuinely still missing plus their wiring, and applies them to current `main`. Net result is +18260/-92 across 55 files, and about 11k of those insertions are the design documents.

## What is here

**Backend** (`a7186b5`) — `ExtensionRecord`, `NodeExtensionRow` and `ArtifactExtensionBundle` in the store with their gorm implementations and adapters; `ExtensionBuilder`; the extension REST handlers; artifact `bundle-extensions` / `bundle-resolve`; node-extension tracking driven from both the REST and WebSocket command-status paths; and `auroraboot sysext --include-path`.

**UI** (`c68e8d0`) — the Extensions list / builder / detail pages, the nav entry, `InstallExtensionDialog`, `HierarchyChipInput`, `ExtensionTypeChip`, and the integrations into ArtifactBuilder, NodeDetail and CommandDialog.

**Docs** (`775b5cc`) — the phased implementation plans and the UI design spec, unchanged.

Extension routes register only when an `ExtensionBuilder` is wired, so a deployment without one keeps exactly today's API surface.

## Changes made against the original branch

These are deliberate, not accidental drift:

1. **`react-router`, not `react-router-dom`.** The three new pages imported from `react-router-dom`, which `main` no longer has.
2. **`kairos/v4/sdk` import paths kept.** The branch predates the monorepo move (b60bb9c); `internal/cmd/sysext.go` keeps `main`'s paths.
3. **CommandDialog bundle state is derived, not cleared in an effect.** The original cleared `bundle`/`bundlePicks` synchronously inside a `useEffect`, which `react-hooks/set-state-in-effect` rejects as an error. The resolved bundle is now stored with the artifact id it belongs to and derived at render. This also fixes a real glitch: previously the prior artifact's extensions stayed on screen while the new resolve was in flight.
4. **`InstallExtensionDialog` re-syncs during render.** Same lint rule; switched to the `prevOpen` pattern `CommandDialog` already uses in this repo.
5. **Two `any` casts replaced** with narrowed types in `api/client.ts` and `ExtensionDetail.tsx`.
6. **Fixed a flaky test in `extension_test.go`.** `Build()` detaches a goroutine that keeps writing under the builder's directory after the spec returns, so `GinkgoT().TempDir()` cleanup raced it and failed with `directory not empty` (reproduced on roughly 1 run in 3). Builder specs now use a scratch dir with best-effort cleanup, and the `List` spec waits for terminal phases rather than for the records to merely exist. 6/6 green after.
7. **Swagger regenerated** via `swag init`, picking up the new endpoints.

## Verification

- `go build ./...` clean, `go vet ./...` clean.
- Go unit tests pass across `./pkg/...` and `./internal/...`, except the 4 pre-existing `internal/cmd` redfish vendor failures, which fail identically on `main` (37 pass / 4 fail on `main`; 39 pass / 4 fail here, the two extra being the new `sysext` specs).
- UI: `tsc --noEmit` clean, `npm run build` clean, `npm test` 48/48, and `npm run lint` back at `main`'s baseline of 0 errors / 3 pre-existing warnings.
- `gofmt` clean on every file this PR touches, other than `pkg/handlers/artifacts.go` and `artifacts_test.go`, which are already dirty on `main`.

## Not in scope

The agent-side half. The dependency this branch declared, kairos-io/kairos-agent#1226, was **closed without merging**, and no replacement exists in `kairos-agent` or `kairos`. The server can build extensions, bundle them onto artifacts and record what a node reports, but nothing installs them on a node until that half lands. #526 should stay open, or a fresh issue should track it, until then.

#526 can be closed in favour of this once you are happy with it, @mudler.

Part of kairos-io/kairos#4298


itxaka-agent:
> any action taken here.

Starting a review. This is a large diff (58 files, +18260/-92) re-applying the extensions work from #526 onto current `main`, so the read will take a while: the diff itself, both referenced PRs, the commit history, and the currently-failing `unit-tests` CI job.

ui/src/components/InstallExtensionDialog.tsx
```diff
@@ -0,0 +1,317 @@
...
+
+  const args: Record<string, string> = {
+    type: extension.type,
+    action,
+    name: extension.name,
+    bootState,
+    now: now ? "true" : "false",
+  };
+  if (action === "install" && extension.rawFilename) {
+    args.source =
```
itxaka-agent, line 79: This sends the admin password to every targeted node. `extensionDownloadUrl` (ui/src/api/extensions.ts:89-95) appends `?token=` read from `localStorage.auroraboot_token`, which is the admin password stored by `login` (ui/src/api/client.ts:32-34). The resulting URL becomes the `source` arg of the `extension` command, so the admin credential is persisted in the commands table, pushed over the WebSocket to every node in the selector, returned to any node polling its own commands, and rendered in this dialog's JSON preview at line 256. pkg/auth/middleware.go:144-147 already states the repo's rule that a node credential must never travel in a URL because of access logs, proxies and Referer headers; this is the admin credential, on lower-trust fleet hardware.

Build the `source` URL without any token. Add a token-free `extensionDownloadPath(id, filename)` to ui/src/api/extensions.ts (same `encodeURIComponent` treatment, no `?token=`) and use it here, keeping `extensionDownloadUrl` for browser anchors only. The node does not need a credential in the URL: it authenticates the download with its own node API key in the `Authorization` header, which `AgentOrAdminMiddleware` already accepts. Apply the same change to e2e/extensions_web_test.go:502-503, which embeds `webAdminPassword` in the same arg.

pkg/server/server.go
```diff
@@ -319,6 +353,13 @@ func New(cfg Config) *echo.Echo {
 	// command actually assigned this artifact to — never an arbitrary artifact.
 	e.GET("/api/v1/artifacts/:id/image", artifactHandler.ExportImage,
 		auth.ArtifactImageMiddleware(cfg.AdminPassword, cfg.NodeStore, cfg.CommandStore))
+	// Extension downloads: admin OR any authenticated node. Nodes need to fetch
+	// extensions bundled into an assigned upgrade command; per-command scoping
+	// analogous to ArtifactImageMiddleware is a follow-up.
+	if extensionHandler != nil {
+		e.GET("/api/v1/extensions/:id/download/:filename", extensionHandler.Download,
```
itxaka-agent, line 360: The extension download link in the UI always returns 401. This route is guarded by `auth.AgentOrAdminMiddleware`, which reads the token only from the `Authorization` header and 401s when it is empty (pkg/auth/middleware.go:112-114) -- it never looks at `?token=`. But the comment above this line and the handler doc at pkg/handlers/extensions.go:332-337 both claim `?token=` is accepted, and ui/src/api/extensions.ts:94 builds exactly that URL for the plain <a href download> anchor at ui/src/pages/ExtensionDetail.tsx:125-130, which cannot set headers. Clicking "Download .raw" yields {"error":"unauthorized"}. No test catches it: extensions_test.go:355 builds the request in-process, and e2e/extensions_web_test.go:207-210 sets the header (line 107).

Accept the admin token from `?token=` on this route, the way the sibling artifact-image download already does. Mirror `ArtifactImageMiddleware` (pkg/auth/middleware.go:148-...): admin may authenticate via header or `?token=`, a node API key is accepted from the `Authorization` header only. Either give the extension download its own middleware with that split, or extend `AgentOrAdminMiddleware` with an admin-only query-token branch and update its doc comment, which currently scopes it to the two shared command routes. Add a route-level spec that GETs the download URL with `?token=<admin>` and no header, and asserts 200.

pkg/handlers/extensions.go
```diff
@@ -0,0 +1,427 @@
...
+	}
+	if req.Name == "" {
+		return c.JSON(http.StatusBadRequest, map[string]string{"error": "name is required"})
+	}
+
+	if err := validateExtensionRequest(req); err != nil {
+		return c.JSON(http.StatusBadRequest, map[string]string{"error": err.Error()})
+	}
+
+	normalized, err := validateHierarchies(req.Hierarchies)
```
itxaka-agent, line 104: A `type=confext` request with a non-empty `hierarchies` array always fails the build. `validateHierarchies` ignores `req.Type`, so the paths are accepted and persisted; the builder forwards them unconditionally as `IncludePaths` (internal/builder/auroraboot/extension.go:245) and `DefaultAurorabootCLIFunc` emits `--include-path` for every type (extension.go:139-141, one line above its own `ServiceReload && Type == "sysext"` guard). But `ConfextCmd.Flags = commonFlagsSysextConfext()` (internal/cmd/sysext.go:59) and that helper never declares `--include-path` -- only `SysextCmd` adds it (sysext.go:44-47). urfave/cli rejects the undefined flag, so the build ends in phase Error with `auroraboot confext: exit status 1`. Even with the flag declared the paths would be dropped: sysext.go:150 gates the allowlist extension on `buildType == "sysext"` and lines 169-172 then overwrite `allowList` for confext. `serviceReload` has the same accepted-then-ignored shape. The UI is unaffected (ExtensionBuilder.tsx:79-80 sends both only for sysext), so this is reachable through the REST API and swagger only. extensions_test.go:121 covers the nil case; the non-nil case is untested.

[Builder-verified: running `auroraboot confext --arch amd64 --output <dir> --include-path /srv name alpine:3.21` exits non-zero with `flag provided but not defined: -include-path`, confirming ConfextCmd does not declare that flag.]

Reject the combination in the handler with a 400 before calling the builder, matching the `hierarchies` field's own documented contract ("sysext-only; /usr implicit" at pkg/builder/extension.go:39). Do the same for `serviceReload` when `type == "confext"`, or coerce it to false the way the UI does. Add a spec asserting 400 for {type: "confext", hierarchies: ["/srv"]}. Alternatively, if confext include paths are wanted later, move `--include-path` into `commonFlagsSysextConfext()` and extend the confext allowlist branch at internal/cmd/sysext.go:169 -- but that is a separate change, not needed to unblock this PR.

```suggestion
if req.Type == "confext" && len(req.Hierarchies) > 0 {
		return c.JSON(http.StatusBadRequest, map[string]string{
			"error": "hierarchies are sysext-only; /etc is the implicit confext hierarchy",
		})
	}

	normalized, err := validateHierarchies(req.Hierarchies)
```

pkg/handlers/extensions.go
```diff
@@ -0,0 +1,427 @@
...
+	status, err := h.builder.Build(ctx, opts)
+	if err != nil {
+		return c.JSON(http.StatusInternalServerError, map[string]string{"error": "failed to start build"})
+	}
+	// Builder.Build synchronously created the ExtensionRecord but doesn't know
+	// about the original SigningKeySetID — only the resolved file paths it
+	// received. Mirror what the artifact handler does: write the handler-only
+	// fields after the builder's initial Save so the persisted record carries
+	// the keyset linkage operators expect on GET.
+	if req.SigningKeySetID != "" && h.store != nil {
```
itxaka-agent, line 151: `signingKeySetId` can be silently lost on every signed build. `ExtensionStore.Create` is a full-row GORM `Save` (internal/store/gorm/store.go:759), and two goroutines do get-then-Save on the same row. `ExtensionBuilder.Build` starts `go b.run(...)` before returning (internal/builder/auroraboot/extension.go:211), and `run`'s first act is `setPhase(BuildBuilding)` -- GetByID then Save of every column (extension.go:345-358). Meanwhile this handler block, running after `Build` returned, does its own GetByID and Save to attach the keyset. If the goroutine's read happens before this Save and its write lands after, `signing_key_set_id` goes back to "" permanently, since later `setPhase` calls re-read the already-blanked row. That is exactly the linkage the comment on line 147-150 says operators expect on GET.

Stop writing the row twice. Add `SigningKeySetID string` to `builder.ExtensionBuildOptions` (pkg/builder/extension.go:31-41), set it from `req.SigningKeySetID` next to the `Signing` paths in this handler, and have the builder write it into the initial `ExtensionRecord` at internal/builder/auroraboot/extension.go:205 -- that Create is synchronous, before the goroutine starts, so there is no window. Then delete this post-Build read-modify-write block. A narrower alternative is a column-scoped update helper on the store, but threading the field is cleaner and removes the whole class of lost update.

itxaka-agent (changes requested):
> any action taken here.

The backend design, the test seams and the Go coverage are solid. Both things worth checking on a rebase this size came back clean: the AuroraBoot half of this feature is self-contained (command dispatch takes an opaque command string, the only node-facing surface is one HTTP download route, and the agent-flow e2e tier is skip-gated behind EXTENSIONS_AGENT_IMAGE), so the closed kairos-agent#1226 does not block this PR. The unit-tests CI failure is a pre-existing timeout in internal/cmd/start-pixie_test.go, unrelated to this diff -- nothing here touches that code path.

Five findings, two blocking, inline below on the files this diff touches. A sixth issue can't be attached inline because it is about a file this diff does NOT touch: `ui/src/lib/buildConfig.ts`'s `PHONEHOME_DESTRUCTIVE_COMMANDS` list is missing an `extension` entry, so `ArtifactBuilder.tsx`'s AllowedCommandsPicker (lines 571-578) never renders a checkbox for it and no artifact built through the UI can ever grant a node permission to run the new `extension` command -- `NodeDetail.tsx`'s new Install-extension flow (lines 391-495) will queue the command against nodes that were never allow-listed for it. The author's own plan doc calls this out as Task 5 (`docs/superpowers/plans/2026-05-26-extensions-3b-ui-install-dialog-artifact.md:773-836`, "extend PHONEHOME_DESTRUCTIVE_COMMANDS") and `e2e/extensions_web_test.go:432` hand-writes the allow-list entry in its fixture because nothing in the product produces it -- the rebase dropped that task. Fix: add `"extension"` next to `"apply-cloud-config"` in the destructive group of `ui/src/lib/buildConfig.ts` (around line 30); it belongs in the destructive list, not the safe defaults, since installing a sysext changes what code runs on the node.

itxaka-agent:
> action taken here.

This slot's re-poll found five different awaiting-author reviews (this PR plus `kairos-io/AuroraBoot#770`, `#776`, `#792`, and `kairos-io/kairos#4604`) all newly non-dormant at once — each author had pushed since our round-0 review. The Second Foundation runs one ticket in flight per slot; picking all five in the same pass would violate that. I'm dropping this ticket back to a human/next cycle rather than half-processing it, and continuing with the oldest of the five this slot.

For context, round 0 requested changes on five points: the admin token leaking into a download URL sent to every fleet node, a download route that 401s despite documentation claiming `?token=` works, a missing `extension` entry in `PHONEHOME_DESTRUCTIVE_COMMANDS` that makes the new command unreachable from the UI, an accepted-then-silently-broken `confext`+`hierarchies` combination (builder-confirmed), and a lost-update race on `signingKeySetId`. New commits and a comment have landed since (latest: 2026-09-15T18:08:29Z, about an unrelated CI flake in `start-pixie`). Whether the five findings themselves are addressed hasn't been re-checked — that's exactly the review this escalation is deferring.

Local review branch: `review-repro/771` (workspace clone only, not pushed). No code was written by the Second Foundation on this ticket.

pkg/server/server.go
ci-robbot: Confirmed and fixed in `eb75bb1`: the route now uses its own `ExtensionDownloadMiddleware`, which mirrors `ArtifactImageMiddleware` (admin via header or `?token=`, node key via header only). Added six route-level specs in `pkg/auth/artifact_download_test.go`, including the `?token=<admin>`-with-no-header 200 you asked for.

pkg/handlers/extensions.go
ci-robbot: Confirmed and fixed in `eb75bb1`, taking your suggestion: the handler 400s `confext` + `hierarchies`, and the same for `serviceReload`. The builder also no longer emits either flag for a confext, as defence in depth. To make that assertable without an exec I pulled the argv out into `AurorabootCLIArgv`, so a spec can check which flags are emitted per type.

pkg/handlers/extensions.go
ci-robbot: Confirmed and fixed in `eb75bb1`, the way you suggested: `SigningKeySetID` is now on `ExtensionBuildOptions` and the builder writes it into the record its synchronous `Create` persists, so the post-Build read-modify-write is gone. Specs cover both ends: the handler passes the ID through, and the builder's row carries it and keeps it across every later phase upsert.

ui/src/components/InstallExtensionDialog.tsx
ci-robbot: The leak is real, but the suggested fix breaks every install, so I have left this one open rather than ship it.

The node cannot authenticate this download over a header. `source` is handed to `kairos-agent sysext install`, which reaches `httpSource.Download` -> `cfg.Client.GetURL` (`agent/pkg/implementations/http/client.go:40`) -> `grab.NewRequest`. There is not one `Header.Set` on that path, so a token-free URL 401s on every targeted node.

`AgentOrAdminMiddleware` accepting node keys from the header is not enough on its own: nothing on the agent side ever puts one there. The same mismatch already exists on `main`, independent of this PR: `downloadArtifact` (kairos `agent/internal/phonehome/handlers.go:244`) builds `/artifacts/<id>/image?token=<node apiKey>` with no header, while `ArtifactImageMiddleware` accepts a node key from the header only, so node artifact pulls should be 401ing today. Filing that separately.

Two designs that keep the fleet working and stop the admin password travelling:

1. Make `source` an opaque `extension:<id>`, resolved agent-side like `artifact:<id>` already is. Matches the existing pattern, but needs a kairos-side change and does not help already-deployed agents.
2. Mint a per-extension download token on the record and put that in the URL, the way `ArtifactRecord.UploadToken` already works for uploads. Self-contained in AuroraBoot, no agent change, but it adds a column.

I lean towards 2 for this PR and 1 as the longer-term shape. Which do you want?

ci-robbot: Fixed in `82f4379`, taking option 2 as I proposed. No answer came back on the choice, and the leak is worth more than the wait; option 1 (`extension:<id>` resolved agent-side) still looks like the right long-term shape, and it needs a kairos-side change to get there.

`ExtensionRecord` grows a `DownloadToken`, minted in the handler and threaded through `ExtensionBuildOptions` so the builder writes it into the row its synchronous `Create` persists, for the same reason the keyset had to move there. `ExtensionDownloadMiddleware` accepts it from `?token=`, scoped to the `:id` in the path, so one extension's token does not open another's. An empty stored token never authorizes: `secureCompare` is a constant-time byte compare, so `""` equals `""`, and any Authorization header gets a bare `?token=` past the both-empty early return and down to that comparison.

Your `extensionDownloadPath` suggestion is in, as the shared builder. `extensionDownloadUrl` keeps the admin token for the browser anchor only; `extensionSourceUrl` is what the Install dialog uses, and it returns null rather than falling back to an admin credential, so the install button disables instead of leaking.

One thing your comment did not cover: `ResolveBundle` (pkg/handlers/artifacts.go) built the same URL server-side with no credential at all, so bundled extensions 401'd on the node rather than leaking. Same token there now.

Specs: five route-level ones in `pkg/auth/artifact_download_test.go`, handler and builder specs for the threading, and the two e2e sysext specs now assert the admin password appears in neither source URL. The `ArtifactImageMiddleware` mismatch I mentioned on `main` turned out to be already filed as kairos-io/kairos#4703, and already fixed: `downloadArtifact` now sends the node key in the Authorization header.

mudler (approved):

mudler:
@ci-robbot rebase is needed

PR #772, opened 2026-09-03
fix(server): rate-limit node-driven endpoints, exempt admin

wrkode:
Adds per-identity rate limiting to the node-driven fleet endpoints, addressing the rate-limiting line item of the fleet-server hardening epic (kairos-io/kairos#4117). **On by default**, generous limits, fully configurable and disable-able.

## What gets limited (and how it's keyed)

| Endpoint | Auth | Limiter key | Rationale |
|---|---|---|---|
| `POST /nodes/register` | reg token (body) | **client IP** (before the token check) | throttles a leaked/guessed-token flood *and* token brute-force |
| `POST /nodes/:id/heartbeat` | node API key | **node ID** | caps a runaway / compromised node |
| `GET /nodes/:id/commands`, `PUT .../status` | node **or** admin | **node ID** (admin skipped) | node polling capped; admin exempt |

## The admin / CAPI exemption is structural

Admin requests set **no** node ID in context (verified in `pkg/auth/middleware.go`). The node limiter's `Skipper` is exactly "skip when `AuthNodeID(c) == ""`", so every admin-bearer call passes straight through. All admin-only routes (`adminGroup`: claim, node-reads, command-create, release) get **no** limiter at all. The CAPI infra provider authenticates as admin and only ever calls those + the shared command poll — so it is never throttled, and the fleet contract is unchanged. An end-to-end server test asserts this (admin fires 10× the node burst on `/commands` and never sees a 429, while a node does).

The limiters are wired **after** the node-auth middleware so the node ID is in context; the registration limiter is wired **before** `RegistrationTokenAuth` so invalid-token attempts are throttled too.

## Configuration (on by default)

Zero config → generous defaults (node 5 req/s burst 20; register ~30/min burst 20 per IP). Overridable:

- `--node-rate-limit <rps>` / `AURORABOOT_NODE_RATE_LIMIT`
- `--register-rate-limit <rps>` / `AURORABOOT_REGISTER_RATE_LIMIT`
- `--disable-rate-limit` / `AURORABOOT_DISABLE_RATE_LIMIT`

## Notes for reviewers

- Built on echo's tested token-bucket (`middleware.RateLimiterWithConfig` + in-memory per-key store), not a hand-rolled limiter.
- **Per-IP key trust:** the registration key is echo's `RealIP` (honours `X-Forwarded-For` behind a trusted proxy). A client that sets its own XFF could rotate the key, so the per-IP limiter is a **flood speed-bump, not the access control** — the registration token is. This is called out in the code comment and the README. A NAT'd rack shares one bucket; raise `--register-rate-limit` or disable for that deployment.
- Tests: middleware unit tests (per-node / per-IP keying, admin exemption, disabled passthrough) + server integration tests (429 fires end-to-end on node commands & registration; admin never throttled).

Refs: kairos-io/kairos#4117

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,8 +71,32 @@ If you reach the UI from another machine (not `localhost`), set `AURORABOOT_URL`
 | `--url https://…` | External URL of this instance, injected into cloud-configs so nodes know where to phone home |
 | `AURORABOOT_ADMIN_PASSWORD` | Override admin password |
 | `AURORABOOT_REG_TOKEN` | Override registration token |
+| `--disable-rate-limit` | Turn off per-identity rate limiting of the node-driven endpoints |
 
 See the full [AuroraBoot reference](https://kairos.io/docs/reference/auroraboot/) for everything else.
+
+### Rate limiting
+
+The node-driven endpoints — registration, heartbeat and command polling — are
+rate-limited **on by default** so a single misbehaving node, or a leaked
+registration token, can't flood the fleet server. Registration is limited per
+client IP; heartbeat and command polling are limited per node. Admin/UI/API
+traffic (including the CAPI infra provider, which authenticates as admin) is
+**never** rate-limited.
+
+The defaults are generous and only ever bite a runaway or a flood. Tune or
+disable them if needed:
+
+- `--node-rate-limit <rps>` / `AURORABOOT_NODE_RATE_LIMIT` — per-node requests/sec
+- `--register-rate-limit <rps>` / `AURORABOOT_REGISTER_RATE_LIMIT` — per-IP requests/sec
+- `--disable-rate-limit` / `AURORABOOT_DISABLE_RATE_LIMIT` — turn it off entirely
+
+If a whole rack of nodes registers at once from behind a single NAT egress IP,
+they share one per-IP registration bucket; raise `--register-rate-limit` or
+disable rate limiting for that deployment. The per-IP key is the client address
+as the server sees it (honouring `X-Forwarded-For` behind a trusted proxy), so
+it's a flood speed-bump rather than a hard boundary — the registration token
+remains the actual access control.
 
 ### Hadron builds
 
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -58,6 +58,27 @@ type Config struct {
 	// goroutines so a server shutdown cancels in-flight Redfish deploys. Defaults
 	// to context.Background().
 	BaseContext context.Context
+
+	// Rate limiting of the node-driven endpoints (registration, heartbeat, command
+	// polling) — fleet-server hardening, kairos-io/kairos#4117. It is on by
+	// default: zero RPS/Burst values fall back to the auth package defaults.
+	// Admin-authenticated requests (the UI and the CAPI infra provider) are never
+	// limited. DisableRateLimit turns the limiters off entirely.
+	DisableRateLimit       bool
+	NodeRateLimitRPS       float64 // per-node requests/sec for heartbeat + command polling
+	NodeRateLimitBurst     int     // per-node burst
+	RegisterRateLimitRPS   float64 // per-IP requests/sec for registration
+	RegisterRateLimitBurst int     // per-IP burst
+}
+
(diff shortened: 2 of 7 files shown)


pkg/server/server.go
```diff
@@ -216,6 +253,9 @@ func New(cfg Config) *echo.Echo {
 	// across nodes.
 	sharedCmd := e.Group("/api/v1/nodes/:nodeID")
 	sharedCmd.Use(auth.AgentOrAdminMiddleware(cfg.AdminPassword, cfg.NodeStore))
+	// After auth, so it limits node polls but skips admin (which sets no node ID)
+	// — keeping the CAPI infra provider's command polling exempt.
+	sharedCmd.Use(auth.NodeRateLimiter(nodeRateRPS, cfg.NodeRateLimitBurst))
```
jimmykarily: `NodeRateLimiter` is called twice (here and on `agentGroup` for heartbeat), and each call builds its own memory store, so a node gets two independent buckets and its effective sustained rate is ~2x `--node-rate-limit`. Intentional, or should the two `Use()` calls share one middleware instance so it's actually one bucket per node?

wrkode: Not intentional — good catch. Fixed: NodeRateLimiter is now built once and the same instance is shared across the heartbeat and shared-command groups, so a node draws both from a single bucket keyed on its node ID (one effective `--node-rate-limit`, not ~2x). Added a regression test that spends the burst with one heartbeat + one command poll and then sees either route throttled. (rebased onto main + pushed)

pkg/server/server_test.go
```diff
@@ -361,3 +361,70 @@ var _ = Describe("Server", func() {
...
+			CommandStore:           &fakeCommandStore{},
+			GroupStore:             &fakeGroupStore{},
+			Builder:                &fakeBuilder{},
+			AdminPassword:          "admin-pass",
+			RegToken:               "reg-token",
+			AuroraBootURL:          "http://localhost:8080",
+			NodeRateLimitRPS:       1,
+			NodeRateLimitBurst:     2,
+			RegisterRateLimitRPS:   0.5,
+			RegisterRateLimitBurst: 2,
```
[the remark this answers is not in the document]
wrkode: Addressed — the wiring tests now use a near-zero rate (0.001 rps) with burst 2, so the bucket refills negligibly during the loop and the 429 is deterministic regardless of CI speed.

README.md
```diff
@@ -71,9 +71,33 @@ If you reach the UI from another machine (not `localhost`), set `AURORABOOT_URL`
...
+registration token, can't flood the fleet server. Registration is limited per
+client IP; heartbeat and command polling are limited per node. Admin/UI/API
+traffic (including the CAPI infra provider, which authenticates as admin) is
+**never** rate-limited.
+
+The defaults are generous and only ever bite a runaway or a flood. Tune or
+disable them if needed:
+
+- `--node-rate-limit <rps>` / `AURORABOOT_NODE_RATE_LIMIT` — per-node requests/sec
+- `--register-rate-limit <rps>` / `AURORABOOT_REGISTER_RATE_LIMIT` — per-IP requests/sec
```
jimmykarily, line 91: We allow the user to override the default value of the rate limit. I wonder if the "burst" setting should somehow follow the user's setting in that case. For example, if the user sets the rate limit to 20, what good does our default burst value of 20 bring? On the other hand, I'm not sure what the relation between the 2 should be, so feel free to leave it as is until we know more.

jimmykarily (approved):

README.md
wrkode: Instinctively, I would say that this is an observe, tune, fine tune, test and repeat. but I'll take a moment to consider and think over it

wrkode: Agreed it's worth thinking about, and thanks for the approve. Leaving burst independent of rate for now, per your suggestion — they're conceptually distinct (burst = instantaneous allowance, rate = sustained refill), and I'd rather not couple them on a guessed heuristic before we see a real case where the fixed default of 20 bites. Easy follow-up to revisit (e.g. defaulting burst as a function of rate) once we have a concrete scenario. Merging now since this is non-blocking.

PR #773, opened 2026-09-03
fix(server): scope node artifact downloads to assigned images

wrkode:
Scopes node artifact downloads to least privilege, closing the last "breaking" item of the fleet-server hardening epic (kairos-io/kairos#4117).

## The hole

`DownloadMiddleware` accepted **any valid node API key** on both `GET /api/v1/artifacts/:id/download/*` and `GET /api/v1/artifacts/:id/image`, with **no per-artifact check**. So a node — or a leaked node API key — could pull **every image the fleet has ever built**, and those images may embed a cloud-config and secrets.

## The one real node-download flow

I mapped the actual flows across the install script, the Go client, the UI, and the upgrade/deploy paths. A node legitimately downloads an artifact in exactly **one** way: an operator queues an `upgrade` / `upgrade-recovery` command whose `Args["source"] == "artifact:<id>"`; the node polls it and pulls that artifact's **container image** via `/image` with its node key. Nothing node-side ever uses `/download/*` (netboot/Redfish read those files from disk server-side). `?token=` is only ever generated with the admin token.

## The fix (least privilege)

| Route | Before | After |
|---|---|---|
| `/download/*` (ISO/UKI/raw/netboot) | admin **or any node key** | **admin-only** (`AdminMiddleware`, still honours `?token=` for UI links) |
| `/image` (container image) | admin **or any node key** | admin **or** a node that an upgrade command actually assigned this artifact to (`ArtifactImageMiddleware`), else **403** |

Authorization keys on the **command assignment**, not the artifact's `TargetGroupID` (that's a build-time cloud-config hint, often empty — not an ACL). It deliberately does **not** filter on command phase: once a node's been told to install an artifact, re-fetching that image (a retry/resumed upgrade) isn't a new exposure — the boundary enforced is *which* artifact, not how many times. Store errors fail closed.

## CAPI-safe

Admin-authenticated callers — the UI and the CAPI infra provider — bypass both checks. When the provider issues an upgrade command to a node, that node's `/image` download is authorized **by the very command the provider issued**. The fleet contract is unchanged.

`DownloadMiddleware` (now unused, and a footgun to leave lying around) is removed.

## Tests

- **Middleware unit tests**: admin (header + `?token=`), assigned node, unassigned node → 403, wrong artifact → 403, non-upgrade command naming the artifact → 403, no/unknown token → 401.
- **Server integration tests**: node key rejected on `/download` (401), scoped on `/image` (403 unassigned; auth passes when assigned or admin) — verified end to end.

Refs: kairos-io/kairos#4117

### DIFF
--- a/pkg/auth/artifact_download_test.go
+++ b/pkg/auth/artifact_download_test.go
@@ -1,0 +1,134 @@
+package auth_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+
+	"github.com/kairos-io/AuroraBoot/pkg/auth"
+	"github.com/kairos-io/AuroraBoot/pkg/store"
+	"github.com/labstack/echo/v4"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+// fakeCommandStore is a minimal store.CommandStore for the authorization tests;
+// only ListByNode is functional.
+type fakeCommandStore struct {
+	byNode map[string][]*store.NodeCommand
+}
+
+func (f *fakeCommandStore) ListByNode(_ context.Context, nodeID string) ([]*store.NodeCommand, error) {
+	return f.byNode[nodeID], nil
+}
+func (f *fakeCommandStore) Create(context.Context, *store.NodeCommand) error { return nil }
+func (f *fakeCommandStore) GetByID(context.Context, string) (*store.NodeCommand, error) {
+	return nil, nil
+}
+func (f *fakeCommandStore) GetPending(context.Context, string) ([]*store.NodeCommand, error) {
+	return nil, nil
+}
+func (f *fakeCommandStore) MarkDelivered(context.Context, []string) error          { return nil }
+func (f *fakeCommandStore) ClaimForDelivery(context.Context, string) (bool, error) { return false, nil }
+func (f *fakeCommandStore) UpdateStatus(context.Context, string, string, string) error {
+	return nil
+}
+func (f *fakeCommandStore) UpdateStatusForNode(context.Context, string, string, string, string) error {
+	return nil
+}
+func (f *fakeCommandStore) Delete(context.Context, string) error         { return nil }
+func (f *fakeCommandStore) DeleteTerminal(context.Context, string) error { return nil }
+
+var _ = Describe("ArtifactImageMiddleware", func() {
+	const (
+		adminPass = "admin-pass"
+		artID     = "art-1"
+	)
+
+	var (
+		e  *echo.Echo
+		ns *fakeNodeStore
+		cs *fakeCommandStore
+		mw echo.MiddlewareFunc
+	)
+
+	BeforeEach(func() {
+		e = echo.New()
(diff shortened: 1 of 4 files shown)


wrkode:
Addressed the Copilot review (node keys via `?token=`): `ArtifactImageMiddleware` now accepts a **node API key only from the `Authorization` header**, never from `?token=`. A node credential in a URL would leak through access logs, proxies, browser history and Referer headers — and no node flow supplies its key that way anyway. The `?token=` query param remains accepted **only for the admin password** (the UI's browser download links; the CAPI infra provider uses the header). Updated the unit test so a node key via `?token=` is rejected (401) even when the node is assigned the artifact. Also gave the server-side test a minimal fake ArtifactStore so `/image` returns a clean 404 instead of tripping the recover middleware. (df40375)

mauromorales:
@mauro-agent please review

mauro-agent:
⏳ **Work claimed by `mauro-agent`**, an AI agent operated by `@mauromorales`, at 2026-09-04T11:55:36Z. If you are another lane about to start substantial work here, please check for a follow-up comment before duplicating it -- this marker exists specifically to prevent that (see mauromorales/mission-control#589).

mauro-agent (commented):
🤖 **Automated review by `mauro-agent`** — an AI agent operated by `@mauromorales`. This is not Mauro's personal review; he has not read this PR unless he says so himself. If you want a human, feel free to ping him directly.

Static review only. I read the diff and the surrounding code in this repository. I did not build this, run the tests, or run any command from the PR.

**Scope**: `pkg/auth/middleware.go`, `pkg/server/server.go`, and the two test files.

**What the change does**: splits the old combined `DownloadMiddleware` into two paths. Raw file downloads (`/download/*`) become admin-only. The container image route (`/image`) accepts an admin, or a node whose command queue (`ListByNode`) holds an `upgrade` or `upgrade-recovery` command naming this artifact as its source.

**What I checked**: I traced the new logic in `ArtifactImageMiddleware` against `AdminMiddleware`, `NodeAPIKeyMiddleware`, and the `CommandStore` interface (all pre-existing) to confirm it matches the stated behavior. It does. The admin path checks both the header and `?token=`. The node path requires the header only, and correctly 401s a node key sent via `?token=`. `nodeAssignedArtifact` fails closed on a store error and on an empty node or artifact ID. The middleware and the `ExportImage` handler both read the artifact ID from the same `:id` path param, so there is no ID mismatch between the check and the download it gates.

**One point worth a second look, not a defect**: `nodeAssignedArtifact` does not check `NodeCommand.Phase` or `ExpiresAt`. The PR body says this is deliberate ("the boundary enforced is which artifact, not how many times"), so a node keeps access to an image from any upgrade command it was ever assigned, including one long completed, failed, or past its `ExpiresAt`. That may be exactly the intended tradeoff, since re-fetching an already-assigned image is not a new exposure. It does mean there is no way to revoke one node's access to one image short of deleting the command row.

Test coverage looks solid: the new middleware unit tests cover admin (header and `?token=`), no credentials, an unknown token, both command types, a node key rejected via `?token=`, an unassigned node, and a node assigned a different artifact. The server-level integration tests check the same boundary end to end.

I found no correctness bugs in the diff.

mauromorales (approved):

PR #776, opened 2026-09-04
fix(nodes): fail reset lifecycle after timeout

### LINKED ISSUE #4287: AuroraBoot: reset lifecycle — fail a reset whose node never returns (timeout sweep)
Follow-up from #4255 / kairos-io/AuroraBoot#717.

The post-reboot reset lifecycle (#717) resolves a reset's outcome when the node re-registers and reports its boot state (`autoreset`→in-progress, `active`→done, `passive`/`recovery`→failed). But if a node **never comes back** — bricked, powered off, network-isolated — its `ResetState` stays `pending`/`in-progress` indefinitely and nothing marks it `failed`.

**Ask:** add a timeout. `ManagedNode.ResetRequestedAt` is already stored, so this can be either a periodic background sweep or a lazy check: when `ResetState` is `pending`/`in-progress` and `ResetRequestedAt` is older than a (configurable) deadline with no re-register since, transition it to `failed` (and optionally flag the node). This lets a CAPI provider stop waiting on a node that will never return.

Sub-issue of #4056.

ci-robbot:
Closes kairos-io/kairos#4287.

## Gap this closes

The post-reboot reset lifecycle (#717) resolves a reset when the node re-registers with a `bootState`. A node that **never comes back** — bricked, powered off, network-isolated — leaves `ResetState` at `pending` or `in-progress` forever, and a CAPI provider that waits on `done|failed` waits forever with it.

## Approach: lazy expiry on read

- New `NodeStore.FailResetBefore(ctx, id, deadline)`. Conditional UPDATE against `WHERE reset_state IN (pending, in-progress) AND reset_requested_at <= deadline`, so a concurrent `SetResetPending` that just refreshed the timestamp is not rolled back to `failed` from a stale snapshot. Mirrors the existing CAS idiom (`CASEjectState`, `AdvanceReset`, `ClaimForDelivery`).
- `NodeHandler` calls this on `GET /api/v1/nodes/:id` and every `GET /api/v1/nodes[...]` variant. UI, controller, and CAPI reconcile all poll one of those, so the terminal state lands before anyone waits on it. No process-wide sweeper.
- Reset commands get an `ExpiresAt` matching the same window. `ClaimForDelivery` now gates on `expires_at`, so a node that reappears after the lifecycle failed does not still execute the queued reset — otherwise the failure would be cosmetic.

## Configuration

- `--reset-timeout` on the `web` subcommand, default 30m.
- `AURORABOOT_RESET_TIMEOUT` env var.
- Negative disables expiration entirely (opt-out for operators who want the old behaviour).

## Provenance

@mudler already had a local branch (`mudler-agent/fix-reset-timeout`, commit `1e5c068`, 2026-08-09) that implemented this exact design and never opened as a PR. I rebased that commit onto current `main`; it applied cleanly against `79c89d9` and the tree still `gofmt`s clean. Authorship is preserved on the commit; I added my own `Signed-off-by` on top of his so DCO stays green as I carry it forward.

## Test plan

- [x] `gofmt -l` clean.
- [ ] CI: `internal/store/gorm` reset lifecycle suite, `pkg/handlers` node + command tests, `pkg/server` handler test wiring. `sqlite3` needs cgo, so the store tests can only run in CI on this box.
- [ ] Manual: `curl -H "Authorization: Bearer <admin>" -X POST /api/v1/nodes/:id/commands -d '{"command":"reset"}'`, then observe `ResetState` on `GET /api/v1/nodes/:id` after the deadline elapses without a re-register.


itxaka-agent:
> any action taken here.

Picking this up for review. Will read the diff, the linked issue (kairos-io/kairos#4287), and the commit history, pull the branch locally, and post findings as inline comments.

pkg/handlers/reset_timeout.go
```diff
@@ -0,0 +1,20 @@
...
+// A negative timeout disables expiration; zero is normalized by the handler's
+// configuration before this helper is called.
+func resetExpired(node *store.ManagedNode, timeout time.Duration, now time.Time) bool {
+	if timeout < 0 || node.ResetRequestedAt == nil {
+		return false
+	}
+	if node.ResetState != store.ResetStatePending && node.ResetState != store.ResetStateInProgress {
+		return false
+	}
+	return !now.Before(node.ResetRequestedAt.Add(timeout))
```
itxaka-agent, line 19: The deadline is computed from node.ResetRequestedAt, and nothing refreshes that column after the reset is issued. SetResetPending in internal/store/gorm/store.go:433 is its only writer in the whole tree; resolveReset's autoreset branch at pkg/handlers/nodes.go:128 advances pending to in-progress without restamping it. So a node that re-registers, reports bootState=autoreset, and is actively wiping gets no fresh budget. Once any GET flips it to failed, the node's later re-register with bootState=active hits the early return at pkg/handlers/nodes.go:122 because failed is not an in-flight state, so the reset is reported failed permanently and LastReset is never stamped. Only a brand-new reset clears it. kairos-io/kairos#4287 asked to fail the lifecycle when the timestamp is older than the deadline "with no re-register since"; this predicate never looks at re-registration.

Re-anchor the deadline on the last proof of liveness instead of on the request. The smallest change that keeps the CAS shape: extend AdvanceReset so the pending to in-progress transition also writes reset_requested_at = now (it already conditionally writes last_reset for the done transition, so the mechanism exists), and call it that way from resolveReset's BootStateAutoReset branch. That gives the in-progress phase its own full window and makes the "no re-register since" semantics real, because a re-register is exactly what restamps it. If you would rather not overload reset_requested_at, add the issue's predicate directly: pass LastHeartbeat into resetExpired and return false when the node has been seen after the deadline. Either way, add a store-level test that a node re-registering with autoreset before the deadline is still in-progress after the original window elapses.

pkg/handlers/nodes.go
```diff
@@ -214,6 +262,9 @@ func (h *NodeHandler) List(c echo.Context) error {
 		if err != nil {
 			return c.JSON(http.StatusInternalServerError, map[string]string{"error": "failed to list nodes"})
 		}
+		if err := h.expireResets(ctx, nodes); err != nil {
```
itxaka-agent, line 265: A failed FailResetBefore write now replaces the entire node list with a 500. GET /api/v1/nodes is the UI's node page and the CAPI provider's read path, so one failed bookkeeping update hides the whole fleet. This contradicts the two nearest precedents: resolveReset at pkg/handlers/nodes.go:121 discards store errors on purpose ("Best-effort: a store error is ignored"), and expireReset itself returns 200 with a stale pending state when the CAS loses at pkg/handlers/nodes.go:149. Both the write-error case and the CAS-loss case leave the same stale ResetState in the response, so failing the request for one and not the other is arbitrary.

Make expiry best-effort like resolveReset. Drop the error from expireResets and expireReset (or ignore it at the call sites with an explicit discard plus a log line), and return the 200 with whatever state the snapshot holds; the next read retries the transition. All four call sites change together: pkg/handlers/nodes.go:265, :281, :291 and the single-node :314. Then replace the "returns an error when the expiry transition cannot be persisted" case in pkg/handlers/nodes_reset_test.go so it asserts a 200 with the node still listed as pending, which is the behavior you want an operator to get when the DB rejects a write.

internal/cmd/web.go
```diff
@@ -109,6 +109,7 @@ var WebCMD = cli.Command{
 		&cli.BoolFlag{Name: "disable-rate-limit", Usage: "Disable per-identity rate limiting of the node-driven endpoints (registration, heartbeat, command polling). Admin/UI/API traffic is never rate-limited regardless. Consider this for a large fleet behind a shared NAT egress IP", EnvVars: []string{"AURORABOOT_DISABLE_RATE_LIMIT"}},
 		&cli.Float64Flag{Name: "node-rate-limit", Usage: "Per-node requests/sec for heartbeat and command polling (0 = generous default). Admin traffic is exempt", EnvVars: []string{"AURORABOOT_NODE_RATE_LIMIT"}},
 		&cli.Float64Flag{Name: "register-rate-limit", Usage: "Per-client-IP registration requests/sec (0 = generous default)", EnvVars: []string{"AURORABOOT_REGISTER_RATE_LIMIT"}},
+		&cli.DurationFlag{Name: "reset-timeout", Value: handlers.DefaultResetTimeout, Usage: "Fail pending or in-progress resets that have not returned within this duration; set a negative duration to disable", EnvVars: []string{"AURORABOOT_RESET_TIMEOUT"}},
```
itxaka-agent, line 112: The new --reset-timeout flag and AURORABOOT_RESET_TIMEOUT env var are documented only in the flag's Usage string. Every comparable web-server knob is in README.md: the "Common settings" table at README.md:66-74 lists --disable-rate-limit among others, and README.md:90-92 lists each rate-limit flag with its env var. An operator reading the README has no way to learn that a 30-minute default exists or that a reset can be marked failed because of it.

Add a row to the "Common settings" table in README.md after line 74, in the same two-column format: the flag and env var in the left cell, and a one-line description in the right cell naming the 30-minute default and that a negative duration disables expiry. Suggested text: | `--reset-timeout 30m` / `AURORABOOT_RESET_TIMEOUT` | How long a node has to come back after a reset before the reset is reported failed; negative disables |.

internal/store/gorm/store.go
```diff
@@ -497,15 +511,16 @@ func (s *Store) MarkDelivered(ctx context.Context, ids []string) error {
...
+// RowsAffected check makes the claim a race-free compare-and-set: when a WS push
+// and an agent poll (or two concurrent polls) both target the same command,
+// exactly one UPDATE matches the still-deliverable row and the loser sees
+// RowsAffected == 0. The expiry predicate is repeated here, rather than relying
+// only on GetPending, so a command cannot expire between selection and claim.
 func (s *Store) ClaimForDelivery(ctx context.Context, id string) (bool, error) {
 	now := time.Now()
 	res := s.db.WithContext(ctx).Model(&store.NodeCommand{}).
-		Where("id = ? AND phase = ?", id, store.CommandPending).
+		Where("id = ? AND phase = ? AND (expires_at IS NULL OR expires_at > ?)", id, store.CommandPending, now).
```
itxaka-agent, line 523: ClaimForDelivery now refuses expired commands, but the fake it is mirrored by does not. fakeCommandStore.ClaimForDelivery at pkg/handlers/fakes_test.go:337 matches on `cmd.ID == id && cmd.Phase == store.CommandPending` and ignores ExpiresAt entirely, so under the fake an expired reset is still claimed and delivered. Any handler-level test of the "a node that reappears after the lifecycle failed must not run the queued reset" path through NodeHandler.GetCommands at pkg/handlers/nodes.go:664 would pass whether or not the gate works. The fakes in this file otherwise take mirroring seriously; fakeCommandStore.GetPending carries a comment explaining that it returns a copy specifically because real gorm GetPending does.

Add the expiry predicate to the fake so it matches the store: skip a command whose ExpiresAt is non-nil and not after time.Now(), returning false. Then add a handler test that a reset command whose ExpiresAt is in the past is absent from the GetCommands response and stays Pending. The gate itself is already covered at the store level by the new case in internal/store/gorm/store_test.go, so this is about keeping the handler-level suite honest rather than about missing coverage of the gate.

itxaka-agent (changes requested):
> any action taken here.

The CAS logic in `FailResetBefore` is correct — it follows the store's existing idiom and the `reset_requested_at <= ?` predicate closes the stale-snapshot race it targets. `ExpiresAt` is threaded correctly through all three command-delivery routes. Both of the two notes the PR author flagged in the description ("neither a defect") check out against the code.

Two things block merge: the expiry deadline is anchored to `ResetRequestedAt` and nothing ever refreshes it on re-registration, so a reset that completes after the window closes is reported `failed` forever even though the linked issue asked to fail only when there has been no re-register since the request; and `expireResets` turns a store write error into a 500 for the whole node list, which is inconsistent with the best-effort handling next to it. Inline comments below name the exact lines and a fix for each, plus a documentation gap and a test-fake mismatch.

Separate from the code: this commit carries `Signed-off-by` from both `mudler` and `ci-robbot` plus an `Assisted-by: Codex` trailer. DCO is satisfied and nothing here is against policy, but the design and the rebase have different authors — worth a maintainer's eye.

pkg/handlers/reset_timeout.go
ci-robbot: Fixed in `4d20f19`, but with a dedicated column rather than by overloading
`reset_requested_at`.

Your diagnosis holds: `SetResetPending` is the only writer, `resolveReset`'s
autoreset branch does not restamp it, and a node mid-wipe therefore shared the
original window and then got stuck in `failed` for good. I did not take the
first suggestion because `ResetRequestedAt` is serialized as `resetRequestedAt`
and documented as "when the reset command was issued", so restamping it would
make the API field lie to a CAPI consumer. `LastHeartbeat` does not work either:
`Register` on an existing node never stamps it (`nodes.go:199-206`), so a
re-register is invisible to it.

So `ManagedNode` gains `ResetProgressAt`, stamped by `AdvanceReset` whenever
`to == in-progress`, which needs no signature change because that transition is
reachable only from a re-register. `resetExpired` measures from the later of the
two timestamps, `FailResetBefore`'s CAS gates on both, and `SetResetPending`
clears it so a new reset starts its own budget.

Store-level tests are there as you asked, and I checked each gate is
load-bearing on its own by reverting it:

- drop the progress predicate from `FailResetBefore` -> "does not fail a reset
  whose node re-registered after the deadline" fails
- drop the `AdvanceReset` stamp -> 3 specs fail
- drop the `SetResetPending` clear -> "clears ResetProgressAt when a new reset
  is requested" fails
- drop the handler-side anchor -> `TestResetExpired/overdue_request,_recent_re-register`
  plus the handler spec fail

That last one needed a `failResetBeforeCalls` counter on the fake: without it
the handler spec passed either way, because the store CAS caught what the anchor
had missed. Each layer is pinned separately now.

pkg/handlers/nodes.go
ci-robbot: Agreed, fixed in `4d20f19`. `expireReset` and `expireResets` no longer return an
error; a store failure is logged and discarded, and all four call sites
(`nodes.go` List x3 and Get) now return the snapshot as-is.

The inconsistency argument is the one that decided it: a lost CAS already
returns a stale `pending` with a 200, so failing the read for a rejected write
gave two different answers to the same user-visible situation.

Replaced the test as suggested, and added the single-node twin, since `Get` had
the same 500.

internal/cmd/web.go
ci-robbot: Added in `4d20f19`, in the "Common settings" table, close to your wording plus a
note that the clock restarts on a re-register (that is the behaviour change in
the same commit, and it is the part an operator reading a 30-minute default
would otherwise guess wrong). Also extended the flag's own `Usage` to say it.

internal/store/gorm/store.go
ci-robbot: Fixed in `4d20f19`, and the fake's `GetPending` had the same gap, so I mirrored
both through one `commandExpired` helper. The real store repeats the predicate
in both places on purpose, so the fake matching only `ClaimForDelivery` would
have left `GetPending` handing out an expired command.

Added the two handler specs you asked for: an expired reset command is absent
from the `GetCommands` response and stays `Pending`, and an unexpired one is
still delivered. Confirmed they are load-bearing by making `commandExpired`
return `false`, which fails the first one alone.

itxaka-agent:
> action taken here.

This slot's re-poll found five different awaiting-author reviews (this PR plus `kairos-io/AuroraBoot#770`, `#771`, `#792`, and `kairos-io/kairos#4604`) all newly non-dormant at once — each author had pushed since our round-0 review. The Second Foundation runs one ticket in flight per slot; picking all five in the same pass would violate that. I'm dropping this ticket back to a human/next cycle rather than half-processing it, and continuing with the oldest of the five this slot.

For context, round 0 requested changes on: the expiry deadline being anchored to `reset_requested_at` with nothing to refresh it on re-registration (so a reset that actually succeeds after the window closes gets reported failed forever), `expireResets` turning a single store-write failure into a 500 for the entire node list instead of the best-effort handling its neighbors use, an undocumented `--reset-timeout` flag/env var, and a test fake that doesn't mirror the store's new expiry gate. Per the author's comment at 2026-09-15T19:20:39Z, all four are addressed in `4d20f19`, with a callout that the reset-timeout semantics changed in a way worth a maintainer's attention. Whether that holds up is exactly the review this escalation is deferring.

Local review branch: `review-repro/776` (workspace clone only, not pushed). No code was written by the Second Foundation on this ticket.

PR #803, opened 2026-09-10
fix(start-pixie): allow empty cloud-config-file argument

mauro-agent:
`netbootmgr.Manager.Start` deliberately passes an empty string for `cloud-config-file` when starting the netboot server (`internal/netbootmgr/manager.go`, see its comment: "Use empty string for cloud-config (not required for netboot)"). `start-pixie`'s own argument validation rejects that same empty string as missing, so every netboot start from the fleet dashboard fails immediately with `all arguments are required`, before anything binds.

`cloud-config-file` only becomes `config_url` in the boot kernel cmdline (`pkg/ops/netboot.go`). An empty one is valid when the booted image doesn't need to fetch a cloud-config at netboot time — exactly the manager's use case.

Fix: drop `cloudConfigFile == ""` from the validation, and reword the error message. Updated the two existing tests whose expected error string changed as a result.

Found and reproduced live: a plain netboot start from the fleet dashboard failed with this error every time; `docker logs`/`docker compose logs` on the running instance showed `all arguments are required` right after each `POST /api/v1/netboot/start`.

### DIFF
--- a/internal/cmd/start-pixie.go
+++ b/internal/cmd/start-pixie.go
@@ -14,10 +14,14 @@ var StartPixieCmd = cli.Command{
 	Name:    "start-pixie",
 	Aliases: []string{"sp"},
 	Usage:   "Start the Pixiecore netboot server and serve custom PXE files (kernel, initrd, squashfs) with a cloud-config.",
-	Description: `Start a Pixiecore-based PXE server to serve a kernel, initrd, and squashfs image for network booting. 
+	Description: `Start a Pixiecore-based PXE server to serve a kernel, initrd, and squashfs image for network booting.
 
 Arguments:
-  cloud-config-file   Path to the cloud-init or cloud-config YAML file.
+  cloud-config-file   Path to the cloud-init or cloud-config YAML file, or ""
+                       to boot without serving one over config_url. An empty
+                       value is only safe when the booted image already has
+                       everything it needs baked in from build time; a fully
+                       unconfigured image will boot but never install itself.
   squashfs-file       Path to the root filesystem squashfs image.
   address             IP address to bind the server (e.g., 0.0.0.0).
   port                Port for the netboot server (e.g., 8080).
@@ -29,8 +33,9 @@ Options:
 
 Example:
   start-pixie user-data.yaml rootfs.squashfs 0.0.0.0 8080 initrd.img vmlinuz --debug
+  start-pixie "" rootfs.squashfs 0.0.0.0 8080 initrd.img vmlinuz --debug
 `,
-	ArgsUsage: "<cloud-config-file> <squashfs-file> <address> <port> <initrd-file> <kernel-file>",
+	ArgsUsage: "<cloud-config-file|\"\"> <squashfs-file> <address> <port> <initrd-file> <kernel-file>",
 	Flags: []cli.Flag{
 		&cli.BoolFlag{
 			Name:  "debug",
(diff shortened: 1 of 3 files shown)


internal/cmd/start-pixie.go
```diff
@@ -46,11 +46,15 @@ Example:
...
 		kernelFile := c.Args().Get(5)
 
-		// Simple argument validation
-		if cloudConfigFile == "" || squashFSfile == "" || address == "" || netbootPort == "" || initrdFile == "" || kernelFile == "" {
+		// Simple argument validation. cloudConfigFile is allowed to be empty:
+		// it becomes config_url in the boot cmdline (pkg/ops/netboot.go), and
+		// an empty config_url is valid when the image doesn't need one fetched
+		// at netboot time -- the netboot manager relies on this (see
+		// internal/netbootmgr/manager.go's Start).
+		if squashFSfile == "" || address == "" || netbootPort == "" || initrdFile == "" || kernelFile == "" {
```
mudler, line 59: I wonder if it makes sense to start pixie without cloud config at all. it was here indeed because without cloud config it wouldn't be functional at all, no?

mauro-agent:
Good question, worth answering precisely rather than just asserting it's fine.

There are two separate cloud-config mechanisms here, and this diff only touches one of them:

1. **Build-time embedded config**: what gets attached in the Artifact Builder UI (hostname, users, k0s settings, etc.). It flows through `opts.CloudConfig` → `schema.Config.CloudConfig` (`internal/builder/auroraboot/builder.go`'s `assembleConfig`), and gets baked directly into the ISO/squashfs during artifact generation. It's already present in the artifact regardless of what `start-pixie` does.
2. **Netboot-time `config_url`**: what `start-pixie`'s `cloud-config-file` argument controls. It becomes `config_url=` in the boot kernel cmdline (`pkg/ops/netboot.go`), an optional extra config the booting kernel can fetch over HTTP, layered on top of whatever's already baked in from (1).

So an empty `cloud-config-file` at netboot time doesn't mean "no config at all". It means "no additional config beyond what the artifact already carries". For the normal fleet-dashboard flow (attach config, build, netboot), the artifact is already fully configured by the time `start-pixie` runs, so there is nothing left to fetch. The original validation was checking the wrong layer, which is what broke every plain netboot start from the dashboard.

Where your worry is still valid: someone could build an artifact with no cloud-config attached at build time either, then PXE-boot into something genuinely unconfigured. That is a real gap worth calling out. Happy to add a startup warning when both are empty, or a code comment at the validation site making this two-path distinction explicit, whichever you'd prefer.

internal/cmd/start-pixie.go
mauromorales: reply happened on the main thread not here sorry 🤦 

https://github.com/kairos-io/AuroraBoot/pull/803#issuecomment-5619066448

mauro-agent:
Pushed a follow-up commit addressing the review so far:

**@Copilot** — both addressed:
- `ArgsUsage`/`Description` now document that `cloud-config-file` may be empty, and explain when that's actually safe (image already fully configured at build time).
- Added a regression test asserting an empty `cloud-config-file` alone doesn't trip the "required" validation, using a short-lived context so it can't hang on the real netboot bind.

**@mudler** — your concern is real and I don't think it should just be dismissed: an image with genuinely no cloud-config anywhere (neither attached at build time nor supplied at netboot start) will boot but never install itself, and that's a bad silent failure mode. I didn't turn it into a hard error, since a fully-baked-in artifact legitimately has nothing left to fetch and that's the common case this PR exists for. Instead it now logs a clear warning the moment `start-pixie` runs with an empty `cloud-config-file`, so it shows up right in the netboot logs instead of only being discoverable by watching the machine sit there unconfigured.

Ran the full `internal/cmd` suite locally (42/42 passing) to confirm none of this broke anything else.

mauromorales (approved):

mauromorales (approved):

PR #804, opened 2026-09-10
feat: show node's remote IP, updated on every heartbeat

### LINKED ISSUE #4590: feat: AuroraBoot show machine IP and update it with the heartbeat
I think it'd be nice to present in the node info the IP it has and update it with the heartbeat in case this changes

Image

And I think it's one of those things worth showing in the nodes list too

Image

mauro-agent:
## Summary

Fixes kairos-io/kairos#4590.

Adds a server-observed `RemoteIP` to a node: captured from the connection at
register time and refreshed on every heartbeat (both the REST heartbeat and
the WebSocket heartbeat path), so a DHCP/NAT change is picked up
automatically. This is distinct from the existing `Addresses` field, which is
self-reported by the agent — `RemoteIP` is what the server actually observed
the connection come from, and it is populated even when an agent sends no
addresses at all.

Shown in the node detail "Node Info" card and as a new column in the nodes
list.

## Test plan

- `go build ./...` and `go vet ./pkg/... ./internal/...` clean.
- `go test ./pkg/handlers/... ./pkg/ws/... ./pkg/auth/... ./pkg/server/... ./internal/store/...` (every package this diff touches) all pass.
- `go test ./internal/cmd/...` fails in my sandbox on unrelated specs (a `/tmp` write restriction the sandbox imposes); `internal/cmd` is untouched by this diff.
- Regenerated `docs/swagger.json`/`swagger.yaml`/`docs.go` via `make openapi`.
- `npm run build` (tsc + vite), `npx vitest run` (56/56), `npx eslint` all
  pass on the `ui/` side.

---

An unattended agent opened this pull request. Nobody has reviewed the diff yet.

https://claude.ai/code/session_0181oEMJkcj7xr62r5RgzTNf

### DIFF
--- a/docs/docs.go
+++ b/docs/docs.go
@@ -1989,6 +1989,10 @@ const docTemplate = `{
                 "phase": {
                     "type": "string"
                 },
+                "remoteIP": {
+                    "description": "RemoteIP is the IP the server observed the node connect from at register or\nheartbeat time (kairos-io/kairos#4590), independent of Addresses: the node\ndoes not report it, and it survives even when an agent sends no addresses at\nall (e.g. the WebSocket heartbeat path, or an older agent). It tracks NAT/DHCP\nchanges because every heartbeat re-observes and overwrites it.",
+                    "type": "string"
+                },
                 "resetRequestedAt": {
                     "description": "ResetRequestedAt is when the reset command was issued (ResetState set to\npending); nil when no reset has been requested.",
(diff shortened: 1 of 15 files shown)


mauromorales (approved):

mauromorales:
@ci-robbot please QA this

mudler (approved):

PR #806, opened 2026-09-10
feat: stream PXE/netboot server logs to the UI

### LINKED ISSUE #4596: feat: AuroraBoot show PXE boot logs
When doing a deployment via PXE boot, I'd like to be able to view at the logs to see why it fails

Image

after a few seconds I just see this with no way to debug

Image

mauro-agent:
## Summary

Fixes kairos-io/kairos#4596 in part.

**Scope note up front:** this closes the half of the problem AuroraBoot can
see today. The netboot server's own output (DHCP requests, TFTP/HTTP file
serves, why it fails to hand a node its files) is now captured and streamed
to the UI. What happens on the node *after* it has the kernel/initrd — a
boot or install failure past the PXE handoff — is not visible anywhere
server-side yet (no console/serial capture, no early-boot phone-home); that
would be a materially larger follow-up, not something this PR attempts.

**What changed:**
- `internal/netbootmgr`: the `start-pixie` subprocess's stdout/stderr, which
  previously went straight to the AuroraBoot process's own output and
  nowhere else, is now teed into a bounded (256 KiB) per-session buffer and
  broadcast live, reusing the existing build-log WS pattern rather than
  inventing a new one.
- New `GET /api/v1/netboot/logs` snapshot endpoint (mirrors
  `GET /api/v1/artifacts/:id/logs`).
- New `{"type":"netboot-log","data":{"chunk":...}}` WS envelope (mirrors
  `build-log`; no id needed since there is at most one netboot session at a
  time).
- Deploy dialog's PXE tab: a log pane appears once a session has produced
  output, shows the snapshot, then appends live chunks, and stays visible
  after Stop so a failure is still readable.

## Test plan

- `go build ./...`, `go vet ./pkg/... ./internal/...` clean.
- `go test ./pkg/... ./internal/store/... ./internal/netbootmgr/...` — all
  pass, including new unit tests for the log buffer (bounding, broadcast
  fan-out, nil-sink safety).
- Regenerated `docs/swagger.json`/`swagger.yaml`/`docs.go` via `make openapi`.
- `npm run build` (tsc + vite), `npx vitest run` (58/58, including a new
  test for the log pane's snapshot-then-live-append behavior), `npx eslint`
  all pass on the `ui/` side.

---

An unattended agent opened this pull request. Nobody has reviewed the diff yet.

https://claude.ai/code/session_0181oEMJkcj7xr62r5RgzTNf

### DIFF
--- a/docs/docs.go
+++ b/docs/docs.go
@@ -563,6 +563,31 @@ const docTemplate = `{
                 }
             }
         },
+        "/api/v1/netboot/logs": {
+            "get": {
+                "security": [
+                    {
+                        "AdminBearer": []
+                    }
+                ],
+                "description": "Returns the current (or most recent) PXE/netboot session's captured output as text/plain. For real-time updates subscribe to the UI WebSocket and filter {type:\"netboot-log\"} envelopes.",
+                "produces": [
+                    "text/plain"
+                ],
+                "tags": [
+                    "Netboot"
+                ],
+                "summary": "Get netboot server log snapshot",
+                "responses": {
+                    "200": {
+                        "description": "OK",
+                        "schema": {
+                            "type": "string"
+                        }
+                    }
+                }
+            }
+        },
         "/api/v1/nodes": {
             "get": {
                 "security": [
--- a/docs/swagger.json
+++ b/docs/swagger.json
@@ -556,6 +556,31 @@
                 }
             }
         },
+        "/api/v1/netboot/logs": {
+            "get": {
+                "security": [
+                    {
+                        "AdminBearer": []
+                    }
+                ],
+                "description": "Returns the current (or most recent) PXE/netboot session's captured output as text/plain. For real-time updates subscribe to the UI WebSocket and filter {type:\"netboot-log\"} envelopes.",
+                "produces": [
+                    "text/plain"
+                ],
+                "tags": [
+                    "Netboot"
+                ],
+                "summary": "Get netboot server log snapshot",
+                "responses": {
+                    "200": {
+                        "description": "OK",
+                        "schema": {
+                            "type": "string"
+                        }
+                    }
+                }
+            }
+        },
         "/api/v1/nodes": {
             "get": {
                 "security": [
--- a/docs/swagger.yaml
+++ b/docs/swagger.yaml
@@ -948,6 +948,23 @@ paths:
       summary: Claim a node from a group
       tags:
       - Groups
+  /api/v1/netboot/logs:
+    get:
+      description: Returns the current (or most recent) PXE/netboot session's captured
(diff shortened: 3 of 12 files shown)


mauromorales (approved):

mauromorales:
@mauro-agent address the issues raised by copilot

mauro-agent:
⏳ **Work claimed by `mauro-agent`**, an AI agent operated by `@mauromorales`, at 2026-09-10T11:43:01Z. If you are another lane about to start substantial work here, please check for a follow-up comment before duplicating it -- this marker exists specifically to prevent that (see mauromorales/mission-control#589).

mauro-agent:
Addressed the issues Copilot raised in review:

- Fixed the logWriter doc comment: it forwards raw write chunks, not lines.
- Guarded the log pane with a session epoch, so a stale snapshot fetch from
  before Start (or a resync after WebSocket reconnect) cannot overwrite a
  freshly-cleared pane.
- Added the same reconnect-resync the build-log view already uses
  (ArtifactDetail), so a dropped socket does not silently miss chunks.

UI test suite (58 tests) and `tsc -b` both pass. Go test could not run in
this environment (no `go` toolchain available here); the only Go change is a
doc comment, no behaviour change.

Ready for another Copilot pass.

mauromorales:
@mauro-agent if you check the image in the QA https://raw.githubusercontent.com/ci-forks/AuroraBoot/qa-proofs-806/ui-3-live-pxe-log.png you will see that the logs paste plain text, including the color codes from the terminal, this should be treated like terminal output like it's done in other logs in the system

mauro-agent:
⏳ **Work claimed by `mauro-agent`**, an AI agent operated by `@mauromorales`, at 2026-09-10T19:01:40Z. If you are another lane about to start substantial work here, please check for a follow-up comment before duplicating it -- this marker exists specifically to prevent that (see mauromorales/mission-control#589).

mauromorales:
@mauro-agent fix the conflicts

mauro-agent:
`@mauromorales` conflicts are fixed and pushed (commit 165e078).

Two upstream changes landed on files this PR also touches:
- #807 reports a reachable advertised address instead of the wildcard bind.
- #803 allows an empty cloud-config-file argument to start-pixie.

`NewManager` now takes both the advertised URL and the log sink. `Start` uses the real cloud-config path when one exists, instead of always passing an empty string. Both features work together: the log pane still streams output, and the reported address is the real one, not the wildcard.

`go build ./...` passes. All Go package tests pass except the `e2e` suite, which times out building a UKI ISO in a container; that needs Docker and is unrelated to this change. The UI test suite (69 tests) and `tsc -b && vite build` both pass.

mauromorales (approved):

PR #809, opened 2026-09-10
fix(ui): show the real server URL in the cloud-config preview

mauro-agent:
The Review-step cloud-config preview (`ui/src/lib/cloudConfigPreview.ts`) hardcoded `phonehome.url` as the literal string `"<server-url>"`, regardless of what `AURORABOOT_URL` the running instance actually has configured. Masking `registration_token` makes sense (it's a credential). Masking the plain server URL doesn't, and it made a real `http://` vs `https://` misconfiguration undebuggable from the dashboard — found this the hard way tonight, standing up a new fleet instance: the only way to confirm what actually got baked into a built artifact was inspecting its files directly on disk after the fact.

Passes `window.location.origin` from the call site instead — correct for the common single-URL deployment where the dashboard and the registration endpoint are the same host. Falls back to the old placeholder if a caller can't determine one.

Added two tests: the real URL flows through, and the fallback still works when `serverURL` is empty. `npx vitest run` and `npx tsc --noEmit` both clean.

This is not an agent, Mauro used AI to help write this.

### DIFF
--- a/ui/src/lib/cloudConfigPreview.test.ts
+++ b/ui/src/lib/cloudConfigPreview.test.ts
@@ -7,6 +7,7 @@ import { PHONEHOME_SAFE_DEFAULTS } from "./buildConfig";
 const base: Parameters<typeof buildCloudConfigPreview>[0] = {
   autoInstall: true,
   registerAuroraBoot: true,
+  serverURL: "https://fleet.example.com",
   groupName: "production",
   allowedCommands: [...PHONEHOME_SAFE_DEFAULTS],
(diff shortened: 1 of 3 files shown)


mauromorales (approved):

mauro-agent:
Duplicate of #805, which I found after opening this — it's the more complete fix (shows the real token too, per the original issue's ask, and also fixes phonehome secrets leaking into the clone-flow's editable cloud-config textarea). Closing in favor of it.

PR #818, opened 2026-09-11
fix(netboot): share state between build-triggered and dashboard netboot starts

mauro-agent:
## What

A build with the "Netboot" output enabled auto-starts pixiecore as a build-pipeline step (`deployer/steps.go`'s `StepStartNetboot`), calling `ops.StartPixiecore` directly and in-process. That start never touches `netbootmgr.Manager` — the thing the dashboard's Start/Stop/Status endpoints actually track — so the server it starts is invisible to the UI and cannot be stopped from it. A later manual "Start Netboot" click then collides with the still-running one on the same ports instead of reporting it's already running.

## How this was found

Reproduced provisioning real hardware over netboot tonight. A build with the netboot output finished, silently started serving. Clicking "Start Netboot" in the dashboard an hour later failed with `listen udp 0.0.0.0:69: bind: address already in use` — there was no Stop button to press, no indication in the UI a server was even running, since `GET /api/v1/netboot/status` only reflects `netbootmgr.Manager`'s own state. The only way to recover was noticing the port conflict in the container logs and restarting the whole server.

## Fix

- Extracted `Manager.Start`'s core into `StartWithPaths(artifactID, cloudConfig, squashfs, initrd, kernel string) error`, taking already-resolved paths instead of reconstructing them from an `<artifactsDir>/<artifactID>/netboot/` convention. `Start` becomes a thin wrapper that resolves those conventional paths and delegates.
- Added `netbootmgr.WithManager`/`FromContext` to thread the server's shared `*Manager` through a `context.Context`.
- `Builder` (the server-side artifact builder) gets a `WithNetbootManager` option, matching its existing `WithLogBroadcaster` pattern, and injects the manager into the context right before running the deploy pipeline.
- `StepStartNetboot`'s callback checks `netbootmgr.FromContext(ctx)`: when present (always true server-side now), it routes through `mgr.StartWithPaths(...)`, sharing state with the dashboard. When absent (the CLI, which has no dashboard to keep in sync with), it falls back to the original direct `ops.StartPixiecore` call unchanged.

## Testing

`go build ./...` and `go vet ./...` clean across the whole module. Added `TestWithManagerRoundTrips`, `TestFromContextWithNoManagerSetReturnsNil`, and `TestStartWithPathsSharesStateWithStart` (the last one also reproduces the "already running" guard directly, the exact bug this fixes). Full `internal/netbootmgr`, `deployer`, `internal/builder/...`, and `internal/cmd` suites pass.

This is not an agent, Mauro used AI to help write this.

### DIFF
--- a/deployer/steps.go
+++ b/deployer/steps.go
@@ -9,6 +9,7 @@ import (
 	"strconv"
 
 	"github.com/hashicorp/go-multierror"
+	"github.com/kairos-io/AuroraBoot/internal/netbootmgr"
 	"github.com/kairos-io/AuroraBoot/pkg/constants"
 
 	"github.com/kairos-io/AuroraBoot/pkg/ops"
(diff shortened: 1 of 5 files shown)


mauromorales (approved):

mauromorales:
@mauro-agent fix conflicts

mauro-agent:
⏳ **Work claimed by `mauro-agent`**, an AI agent operated by `@mauromorales`, at 2026-09-18T09:06:20Z. If you are another lane about to start substantial work here, please check for a follow-up comment before duplicating it -- this marker exists specifically to prevent two lanes from independently doing the same work.

mauro-agent:
This is an unattended update: no human read this specific commit before it was pushed.

Rebased on main to clear the conflict `@mauromorales` flagged. Both sides touched `internal/netbootmgr/manager.go` (this branch's `context.Context` carrier vs. main's bounded log buffer) and `internal/cmd/web.go` had a duplicate `netbootManager` declaration once main's log-broadcast wiring landed independently. Kept both features, removed the duplicate, and updated three stale `NewManager(...)` test call sites to the two-argument signature main introduced.

`go build`, `go vet`, and `go test ./internal/netbootmgr/...` all pass (14/14).

PR #821, opened 2026-09-14
fix(ui): name the node version fields after what they hold

### LINKED ISSUE #4617: AuroraBoot artifact version vs kairos version
When building artifacts, you can provide a version, which is useful if you want to maintain your own release cadence. However, I think the way we present that version in AuroraBoot Fleet is confusing.

In the screenshot, the artifact version v1.1.0 is shown as Kairos, while the actual Kairos version is effectively represented by Agent Version (v4.3.0-47-ga83f003c).

Now that Kairos uses a single version across the binaries, I think we should make this distinction clearer.

I’d suggest:

* Agent Version → Kairos Version
* Kairos → Image Version (or Artifact Version)

I slightly prefer Image Version here, since this screen is describing the node and the version of the image it is running.

Image

itxaka-agent:
> action taken here.

Fixes: kairos-io/kairos#4617

## What

The Fleet dashboard's node detail page (`ui/src/pages/NodeDetail.tsx`) had its two version fields labelled the wrong way around. "Agent Version" rendered `node.agentVersion`, which is the Kairos version the node is actually running (e.g. `v4.3.0-47-ga83f003c`). "Kairos" rendered `node.osRelease.KAIROS_VERSION`, which is the version of the image/artifact the node was built from (e.g. `v1.1.0`). @mauromorales's report has the full screenshot and reasoning.

Relabelled per his proposal:

- `Agent Version` -> `Kairos Version`
- `Kairos` -> `Image Version`

Labels only — neither underlying field, nor any API/DTO type, changed.

## Verification

- Added `ui/src/test/nodeDetail.versionLabels.test.tsx`: renders the page with distinct `agentVersion`/`osRelease.KAIROS_VERSION` values and asserts both corrected labels appear next to the right value, and neither old label string survives.
- Full suite: `npm test` in `ui/` — 11 files / 67 tests, green.
- No QEMU/ISO reproduction — this is a UI copy change with no boot-path involvement.
- One non-blocking observation for a maintainer: the node list view (`ui/src/components/NodeTable.tsx`) already calls this same `agentVersion` field "OS Version". After this PR the detail page and list page name it differently ("Kairos Version" vs "OS Version"). The issue only asked about the detail page, so bringing the list column in line felt like scope creep for this PR — worth a follow-up if you'd like the two screens to agree.

Audit summary: https://github.com/kairos-io/kairos/issues/4617#issuecomment-5662718090


itxaka-agent:
> This comment was generated by software. A human maintainer can override.

## test run

`npm test` in `ui/` — 11 files, 67 tests, all green. Screenshot shows the head sha and run timestamp on the same terminal capture:

npm test output

Screenshot blob sha: `88caca60ce18184692fc5770a5c827486060a462` (host repo: `itxaka-agent/triage-assets`, path `AuroraBoot/821/tests-npm.png`). Run against head `53b34c5ce7d17d0a76812ce98f8b98c7d2f1b7cf` at `2026-09-16T17:08:44+02:00`.

PR #825, opened 2026-09-16
fix(secureboot): skip certificates already in the enrolled database

### LINKED ISSUE #2432: enki should skip duplicate keys when importing custom keys
In case the user imported our keys manually before exporting all of the from UEFI, or in case they forgot to skip the microsoft certs and they were already there in the custom keys, it would be good if enki checked what keys have been imported and skipped duplicates.

ci-robbot:
`generateAuthKeys` builds one signature database from up to three sources: the
freshly generated certificate, the Microsoft certs sbctl bundles, and the certs
read out of `--custom-cert-dir`. The last two are merged with
`signature.SignatureDatabase.AppendDatabase`, which is a plain `append` of whole
signature lists with no membership check.

Firmware that trusts Microsoft exports the Microsoft certificates along with
everything else. So a user who exports KEK/db from their firmware and passes the
export to `--custom-cert-dir`, without also passing `--skip-microsoft-certs`,
gets every Microsoft certificate enrolled **twice** in `KEK.esl`/`KEK.auth` and
`db.esl`/`db.auth`. That is the second case in kairos-io/kairos#2432, re-scoped
from enki (archived) to the code that replaced it.

### Why the library's own dedup does not catch it

`certs.GetOEMCerts` stamps every entry it returns with Microsoft's owner GUID,
`certs.GetCustomCerts` stamps every entry it returns with a fixed `custom` GUID,
and `SigDataExists` matches on owner **and** data. Measured against the real
libraries:

```
OEM    owner=77fa9abd-0359-4d32-bd60-28f4e78f784b len=1462
CUSTOM owner=88a69775-5ad7-45d9-9f34-cec43e1f1989 len=1462  sameCert=true
oem.BytesExists(CERT_X509, custom-owner, sameCert) = false
```

What firmware ends up trusting is the certificate, so the certificate is what
has to be unique. `signatureDataPresent` matches on the signature type and the
payload, and ignores the owner.

### No change for anyone not hitting this

`appendDatabaseSkippingDuplicates` filters the incoming lists rather than
rebuilding them, so a merge that drops nothing serializes to exactly the bytes
`AppendDatabase` produced before. That is asserted directly, and it is also what
pins the recomputed `ListSize`. The public API (`GenerateKeySet(Options)`) is
unchanged; only the unexported `generateAuthKeys` gained a logger, so each skip
is reported with the certificate subject.

### Verification

Four tests in `pkg/secureboot/genkey_dedup_test.go`, hermetic (they build key
material in Go rather than shelling out to openssl):

- `SkipsCertsAlreadyEnrolled` (KEK and db) reproduces the report.
- `KeepsDistinctCustomCerts` pins that a genuinely new custom cert is still enrolled.
- `NoCustomCertsIsUnchanged` guards the common path.
- `ByteIdenticalWhenNothingIsDropped` pins the no-regression promise.

Mutation-checked, each mutation applied and reverted:

| mutation | result |
|---|---|
| both merges back to `AppendDatabase` (pre-fix) | every Microsoft cert `appears 2 times, want 1`, in both KEK and db |
| `signatureDataPresent` always true | all four fail, incl. `distinct custom cert appears 0 times` |
| `ListSize` off by one | all four fail |

The two pre-existing `TestGenerateKeySet_*` tests fail on my box only because
`openssl` is not on PATH (`exec: "openssl": executable file not found`); that is
unrelated to this change and CI has it.

### Not changed, noted while here

`prepareCustomDerDir` names each extracted cert `<keyType><serial>`. Serial
numbers are only unique per issuer, so two custom certs from different CAs that
happen to share a serial silently overwrite each other, dropping a legitimate
key. Separate defect from this one, left alone to keep the diff reviewable.

Fixes kairos-io/kairos#2432

> Opened by ci-robbot in an unattended session. No human has read this diff yet.


pkg/secureboot/genkey_dedup_test.go
```diff
@@ -0,0 +1,252 @@
...
+//
+// A user who exports KEK/db from their firmware and passes the export through
+// --custom-cert-dir, without also passing --skip-microsoft-certs, hands us the
+// Microsoft certs twice: once from sbctl's bundled copy and once from their own
+// export. The two arrive under different owner GUIDs (certs.GetOEMCerts tags
+// them "microsoft", certs.GetCustomCerts tags everything "custom"), so the
+// library's own owner+data dedup cannot see it, and SignatureDatabase.
+// AppendDatabase does no membership check at all. The cert must still land in
+// the enrolled database exactly once.
+func TestGenerateAuthKeys_SkipsCertsAlreadyEnrolled(t *testing.T) {
```
jimmykarily: @ci-robbot we use ginkgo/gomega styled tests in kairos-io org. Switch.

ci-robbot: Switched in `378c90b`: suite bootstrap added, the two per-variable cases are `DescribeTable` entries, helpers report through `GinkgoHelper`. Checked the conversion kept its teeth rather than just compiling: putting the naive `AppendDatabase` back fails both dedup entries, and dropping a signature in the merge fails all six specs.

jimmykarily (approved):

PR #756, opened 2026-08-25
feat: materialize catalog extensions

mudler-agent:
## What this draft contains

Adds the focused AuroraBoot materializer for named sysext catalog entries. It uses the shared Kairos SDK resolver, pulls one digest-pinned OCI raw layer, validates the Kairos sysext media type, and writes artifacts atomically.

## Why this remains a draft

The follow-up wiring into classic ISO and UKI ISO build paths is not included. The required in-worktree package verification was blocked because this host has Go 1.26.5 while the repository requires 1.26.6, and module/toolchain downloads return HTTP 403. The package tests passed only in an isolated compatibility module using the exact SDK source, which is not sufficient to mark the task complete.

This draft preserves the reviewed implementation for CI and follow-up work without presenting it as complete.

Part of kairos-io/kairos#4298.

PR #830, opened 2026-09-18
feat(extensions): materialize catalog extensions in ISOs, rebased from #756

ci-robbot:
Rebase of #756 onto current `main`, as @mudler asked on that PR. #756 has been conflicting since its last push on 2026-08-28, so its pending QA could not run.

Two of the three original commits carry over unchanged. The third, `2f1d5e7` ("fix(deps): include the extension catalog SDK"), is dropped: it only pinned `github.com/kairos-io/kairos/v4` to the pseudo-version `v4.2.1-0.20260825071302-240edb85a56f` so that `sdk/extensions` would resolve. `main` is already on the `v4.3.0` tag, which ships `sdk/extensions` with the same `Parse(io.Reader) (Catalog, error)` and `Catalog.Resolve(name, version, arch) (Resolved, error)` signatures the code calls. `go mod tidy` is a no-op on this branch, so no dependency change is needed.

Verification on this branch:

| Check | Result |
| --- | --- |
| `go test ./pkg/extensions ./pkg/schema ./pkg/ops ./deployer` | pass |
| `go test ./pkg/uki` | pass (scratch copy, `go-ukify` stubbed: no C compiler here, and the suite does not touch it) |
| `go test ./internal/cmd` | 44/48, the 4 `redfish deploy` specs also fail identically on `main` in the same environment |
| `go vet` on all six touched packages | clean |

The branch adds 6 specs to `internal/cmd` (42 to 48) and all 6 pass. CI is the real gate for the e2e and ISO jobs.

Closes #756.

Part of kairos-io/kairos#4298


internal/cmd/build-iso.go
```diff
@@ -68,6 +69,14 @@ var BuildISOCmd = cli.Command{
 			Name:  "live-console",
 			Usage: "Replace the console options used when booting from the live/installer ISO",
 		},
+		&cli.StringSliceFlag{
+			Name:  "extension",
+			Usage: "Named system extension to include, optionally with @version (repeatable)",
+		},
+		&cli.StringFlag{
```
mudler: @ci-robbot this should be a list, like we do with the kairos-agent.

pkg/extensions/catalog.go
```diff
@@ -0,0 +1,190 @@
...
+		if !requestPart.MatchString(parts[1]) {
+			return Request{}, fmt.Errorf("invalid extension request %q", value)
+		}
+		request.Version = parts[1]
+	}
+	return request, nil
+}
+
+// Materialize resolves requests and writes their raw OCI layers to destination.
+func Materialize(ctx context.Context, catalogSource string, requests []Request, architecture string, destination string, insecure bool) ([]string, error) {
```
mudler: here again, there can be more catalogs. See already the code because we have helpers

mudler:
@ci-robbot see my comments above

internal/cmd/build-iso.go
ci-robbot: Done: `--extensions-catalog` is now repeatable on both `build-iso` and `build-uki`, and `iso.extensions_catalog` became `iso.extensions_catalogs`.

pkg/extensions/catalog.go
ci-robbot: Switched to the SDK helper: resolution now goes through `extensions.Catalogs`, searched in order with the first catalog publishing a name winning, same as the agent. That helper landed after v4.3.0, so the commit bumps `kairos/v4` to a master pseudo-version.

mudler (approved):

mudler:
@ci-robbot what about the UI changes? we want as well to be able to select available extensions from the catalog during build time. And let's stab as default the hadron one for now which should be able to override

PR #846, opened 2026-09-20
feat(uki): selinux support

### LINKED ISSUE #4633: feat:  selinux (uki)
Follow up from https://github.com/kairos-io/kairos/pull/4519

thebhdn:
Hey armadillos,

Closes https://github.com/kairos-io/kairos/issues/4633

This PR brings selinux support for uki. 

As I understood correctly - currently trusted boot was only tested on ubuntu/fedora/hadron, hence I've only added fedora and if we plan to add more distros switch statement can be extended.


ci-robbot (commented):
Thanks for this, and for the tests. CI is green and the cmdline it builds matches what the GRUB path already emits in `kairos-init/pkg/bundled/bundled.go` (`security=selinux selinux=1 enforcing=0 rd.cos.selinux=${selinux_mode}`), so the shape is right. Four things before this is ready:

**1. A dropped `install.selinux` is completely silent.** `isSelinuxSupported` wraps the whole parse, so on a non-fedora image the build emits `selinux=0` and logs nothing. Worth parsing first and logging when the request is dropped, otherwise the user gets an image that ignores their cloud-config with no clue why. It also makes `parseSelinuxOptions`'s "invalid selinux mode" warning unreachable on every image where it could matter.

**2. The gate is narrower than GRUB's, on a different key.** `bundled.go:431` enables SELinux for `KAIROS_FAMILY` in `redhat` or `suse`; this gates on `KAIROS_FLAVOR == "fedora"`, which is os-release `ID`, so Rocky, AlmaLinux, RHEL and openSUSE are excluded. If that is the intended scope for now, fine, but matching on family would keep the two paths from drifting.

**3. This is a second parser for `install.selinux`.** `agent/internal/agent/hooks/gruboptions.go:SelinuxGrubOpts` already implements enabled/mode plus the default-to-permissive and warn-on-unknown-mode behaviour, and `sdk/types/install.SelinuxOptions` is the type. AuroraBoot already depends on `kairos/v4/sdk`; reusing it avoids the two copies diverging the next time the schema changes.

**4. `--cloud-config` only works before the subcommand.** `build-iso` declares its own `cloud-config` flag (`internal/cmd/build-iso.go:25`), so `auroraboot build-iso --cloud-config cc.yaml ...` works, while `auroraboot build-uki --cloud-config cc.yaml ...` fails with "flag provided but not defined". The new spec asserts the error does *not* contain that string, which passes for any other error too, so it would not catch the regression it is guarding.

Minor: `isSelinuxSupported` logs through the package-global `internal.Log` while `Build` already has a `log` in scope, and the `etc/os-release` fallback cannot work as written since a plain distro os-release has no `FLAVOR` key (the sdk then reads the build host's `/etc/os-release`, not the rootfs).

thebhdn:
Hey @Itxaka, @jimmykarily,

I want to clarify this  mr robot comment:

>The gate is narrower than GRUB's. setSelinux enables SELinux for KAIROS_FAMILY in redhat or suse; isSelinuxSupported matches KAIROS_FLAVOR == "fedora". KAIROS_FLAVOR is os-release ID (kairos-init/pkg/stages/steps_init.go), not a family, so Rocky, Alma, RHEL and openSUSE UKI images silently get selinux=0 while their GRUB siblings get SELinux. Gating on KAIROS_FAMILY would match the existing behaviour and need no per-distro list.

I decided to have conditional for `selinux` based on `FLAVOR` as in trusted boot [docs](https://kairos.io/docs/installation/trustedboot) I see:

>Tested distributions
Commands in this page are written for Hadron, but if adapted they should work on these other distributions: Ubuntu 24.04 and Fedora 40.

We discussed that this is related to systemd versions so question should we keep this check based on `FLAVOR` or follow the same pattern as for grub opts - `KAIROS_FAMILY` where we support opensusre/rhel distros? If yes then maybe add  the same warning, or field in cloud config:

```yaml
install:
    yesIKnowWhatIamDoing: true
```

😆

pkg/uki/uki.go
```diff
@@ -1044,8 +1072,8 @@ func createContainer(sourceDir, outputDir, artifactName, outputName string, log
 // GetUkiCmdline returns the set of boot entries (one per cmdline variant) used
 // to generate UKI EFI files. Extend mode appends to the default cmdline and
 // produces a single entry; extra mode produces one entry per extra cmdline.
-func GetUkiCmdline(cmdlineExtend, bootBranding string, extraCmdlines []string, cmdLinesV2 bool) []utils.BootEntry {
-	defaultCmdLine := constants.UkiCmdline + " " + constants.UkiCmdlineInstall
+func GetUkiCmdline(baseCmdline, cmdlineExtend, bootBranding string, extraCmdlines []string, cmdLinesV2 bool) []utils.BootEntry {
```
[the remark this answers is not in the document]
jimmykarily: I suspect every consumer of this method is either in this repo or the kairos monorepo, so we could just adapt them. An alternative would be to make baseCmdline a memoized function that is called everywhere the old value was used. Looking at the PR's diff, it looks feasible to me but I'm not sure the config is available everywhere (to be passed down to that function). Maybe something to investigate.

itxaka-agent:
> any action taken here.

Picking this up for review — checking the SELinux gating logic, the cloud-config flag wiring, and the existing feedback against the code.

pkg/uki/uki.go
```diff
@@ -294,9 +302,28 @@ func Build(opts Options) (err error) {
 		return err
 	}
 
+	// Build the base cmdline from cloud config: enabled install.selinux
+	// replaces the hardcoded selinux=0, otherwise the base stays
+	// identical to constants.UkiCmdline.
+	baseCmdline := constants.UkiCmdline
+
+	if ok := isSelinuxSupported(sourceDir); ok {
```
itxaka-agent, line 310: When the source image is not redhat/suse family, install.selinux is dropped with no log line at any level. A user who sets install.selinux.enabled: true in the cloud-config and builds a UKI from an Ubuntu image gets a cmdline with selinux=0 and no indication why. The same nesting also means a malformed cloud-config is a hard build failure on a Fedora image (parseSelinuxOptions returns the YAML error) and is never even parsed on an Ubuntu image, so the build succeeds.

Invert the nesting: call parseSelinuxOptions first whenever opts.CloudConfig is non-empty, so YAML errors surface the same way on every image. Then check isSelinuxSupported, and when the config asked for SELinux on an unsupported family emit a warning that names the family (e.g. `log.Warnf("install.selinux.enabled is set but image family %q does not support SELinux; booting with selinux=0", family)`). isSelinuxSupported should return the family string alongside the bool so the warning can name it.

internal/cmd/build-uki.go
```diff
@@ -154,6 +155,15 @@ var BuildUKICmd = cli.Command{
 		}
 		log := logger.NewKairosLogger("auroraboot", logLevel, false)
 
+		cloudConfig := ""
+		if ccPath := ctx.String("cloud-config"); ccPath != "" {
```
itxaka-agent, line 159: Confirmed by builder: a minimal urfave/cli v2.27.7 program with the same flag shape fails with "flag provided but not defined: -cloud-config" when the flag follows the subcommand and succeeds when it precedes it, matching this file's structure exactly. build-uki reads ctx.String("cloud-config") but never declares the flag; it is only defined app-level (internal/cmd/app.go:41), so `auroraboot build-uki --cloud-config cc.yaml oci:img` fails and only the pre-subcommand form works. build-iso declares its own local flag with a -c alias, so the two subcommands disagree on where the flag goes.

Add the flag to BuildUKICmd.Flags, mirroring build-iso: a cli.StringFlag named cloud-config, alias c. Then update internal/cmd/build-uki_test.go:31 to exercise the post-subcommand form (`app.Run([]string{"", "build-uki", "--cloud-config", cc, "some/image:latest"})`).

internal/cmd/build-uki_test.go
```diff
@@ -26,4 +27,15 @@ var _ = Describe("build-uki", Label("uki", "cmd"), func() {
...
 		Expect(err.Error()).ToNot(ContainSubstring("flag provided but not defined"))
 	})
+
+	It("accepts the app-level --cloud-config flag before the subcommand", Label("flags"), func() {
+		cc := GinkgoT().TempDir() + "/cc.yaml"
+		Expect(os.WriteFile(cc, []byte("#cloud-config\nusers:\n  - name: kairos\n"), 0o644)).To(Succeed())
+
+		err = app.Run([]string{"", "--cloud-config", cc, "build-uki", "some/image:latest"})
+
+		Expect(err).ToNot(BeNil())
```
itxaka-agent, line 37: The test only asserts the error string does not contain "flag provided but not defined" or "unknown flag". It passes even if ReadCloudConfig is never called or the parsed config never reaches uki.Options.CloudConfig — any unrelated failure satisfies all three assertions.

Assert the wiring, not the absence of a parser error. Capture uki.Options the way internal/builder/auroraboot/builder_test.go does with capturedUKI and assert CloudConfig equals the file contents, or point --cloud-config at a nonexistent path and assert the error names that path.

pkg/uki/uki.go
```diff
@@ -294,9 +302,28 @@ func Build(opts Options) (err error) {
...
+		if opts.CloudConfig != "" {
+			enabled, selinuxMode, err := parseSelinuxOptions(log, opts.CloudConfig)
+			if err != nil {
+				return err
+			}
+			if enabled {
+				selinuxParams := fmt.Sprintf("security=selinux selinux=1 enforcing=0 rd.cos.selinux=%s", selinuxMode)
+				baseCmdline = strings.Replace(baseCmdline, " selinux=0", " "+selinuxParams, 1)
```
itxaka-agent, line 318: The SELinux fragment is spliced in with strings.Replace(baseCmdline, " selinux=0", " "+selinuxParams, 1). This depends on the exact token " selinux=0" existing in constants.UkiCmdline. If that constant is ever reordered or renamed, Replace returns the string unchanged, baseCmdline keeps selinux=0, and SELinux is silently not enabled while the log still prints "SELinux enabled in cloud-config, mode: ..." on the next line.

Capture the Replace result in a new variable, compare it against the input, and error (or at least log an error) when they are equal. pkg/uki/uki_test.go:236 already guards its own fixture this way — production code should too.

pkg/uki/uki.go
```diff
@@ -1138,3 +1220,33 @@ func FindFirstFileInDir(dir, pattern string) (string, error) {
...
 	return foundFile, nil
 }
+
+func isSelinuxSupported(rootfs string) bool {
+	family, err := sdkutils.OSRelease("FAMILY", filepath.Join(rootfs, "etc/kairos-release"))
+	if err != nil {
+		// fallback to os-release
+		family, err = sdkutils.OSRelease("FAMILY", filepath.Join(rootfs, "etc/os-release"))
+		if err != nil {
+			internal.Log.Logger.Error().Err(err).Msg("failed to get image family")
```
itxaka-agent, line 1230: isSelinuxSupported logs its only error through the package-global internal.Log instead of the logger the caller passed in opts.Logger, which Build already holds as log. internal/builder/auroraboot/builder.go:665-668 tees that caller logger into a dbLogWriter for the web UI, so the "failed to get image family" line never reaches the UI build log — the build just quietly produces selinux=0.

Give isSelinuxSupported a log *logger.KairosLogger first parameter, pass log from the call site at uki.go:310, and replace the internal.Log call with log.Errorf("failed to get image family: %v", err). Drop the now-unused github.com/kairos-io/AuroraBoot/internal import.

pkg/uki/uki.go
```diff
@@ -1138,3 +1220,33 @@ func FindFirstFileInDir(dir, pattern string) (string, error) {
 	}
 	return foundFile, nil
 }
+
+func isSelinuxSupported(rootfs string) bool {
+	family, err := sdkutils.OSRelease("FAMILY", filepath.Join(rootfs, "etc/kairos-release"))
+	if err != nil {
+		// fallback to os-release
+		family, err = sdkutils.OSRelease("FAMILY", filepath.Join(rootfs, "etc/os-release"))
```
itxaka-agent, line 1228: Confirmed by builder: sdk/utils.OSRelease at the pinned pseudo-version returns the build host's real /etc/os-release value when the requested key is absent from the caller-supplied file, verified by feeding it a file containing only FOO=bar and getting the host's PRETTY_NAME back. KAIROS_FAMILY is also written only to /etc/kairos-release by kairos-init, never to os-release, so this fallback cannot fire correctly on a real image — it can only leak the build host's own file.

Read rootfs's etc/kairos-release directly for KAIROS_FAMILY instead of relying on OSRelease's hidden fallback. If you keep an os-release fallback at all, justify it with a real image that has KAIROS_FAMILY there; otherwise drop it along with the pkg/uki/uki_test.go:279-282 entry, which writes KAIROS_FAMILY into an os-release shape Kairos does not produce.

pkg/uki/uki.go
```diff
@@ -294,9 +302,28 @@ func Build(opts Options) (err error) {
...
+			}
+			if enabled {
+				selinuxParams := fmt.Sprintf("security=selinux selinux=1 enforcing=0 rd.cos.selinux=%s", selinuxMode)
+				baseCmdline = strings.Replace(baseCmdline, " selinux=0", " "+selinuxParams, 1)
+				log.Infof("SELinux enabled in cloud-config, mode: %s", selinuxMode)
+			}
+		}
+	}
+
 	entries := append(
```
itxaka-agent, line 324: The SELinux fragment lands in baseCmdline, inherited by every boot entry including the install-mode entry and single-efi entries. bundled.go's GRUB-path setSelinux instead forces selinux=0 on recovery (label != COS_SYSTEM), so UKI install/recovery entries diverge from GRUB and could boot enforcing against a filesystem that was never relabelled.

Exclude the install-mode/recovery entries from the SELinux splice to match GRUB, or state explicitly in the PR why UKI install/recovery is safe with selinux=1 where GRUB's is not.

pkg/uki/uki.go
```diff
@@ -1110,6 +1138,60 @@ func NameFromCmdline(basename, cmdline string) string {
...
+		return false, "", nil
+	}
+
+	mode = strings.TrimSpace(selinux.Mode)
+	switch mode {
+	case "":
+		mode = "permissive"
+	case "permissive", "enforcing":
+	default:
+		log.Warnf("invalid selinux mode %q in cloud-config, falling back to permissive", mode)
```
itxaka-agent, line 1189: An invalid mode produces two warnings for the same input: isSelinuxModeExists (around line 1172) logs "unknown install.selinux.mode ... using permissive" inside the decode loop, then the normalisation switch here logs "invalid selinux mode ... falling back to permissive" again for the same value.

Drop the isSelinuxModeExists warning and let the normalisation switch be the single validation/warning point.

pkg/uki/uki.go
```diff
@@ -933,7 +960,8 @@ func sumFileSizes(filesMap map[string][]string) (int64, error) {
 }
 
 func createImgWithSize(imgFile string, size int64) error {
-	cmd := exec.Command("dd",
+	cmd := exec.Command(
```
itxaka-agent, line 963: exec.Command("dd", is reformatted onto two lines in createImgWithSize. This has nothing to do with SELinux or cloud-config and just makes the diff larger than the change it describes.

Revert this hunk so the function reads exactly as it did before the PR.

```suggestion
	cmd := exec.Command("dd",
```

internal/cmd/build-uki.go
```diff
@@ -154,6 +155,15 @@ var BuildUKICmd = cli.Command{
...
+			cc, err := config.ReadCloudConfig(ccPath, map[string]interface{}{})
+			if err != nil {
+				return err
```
itxaka-agent, line 162: A cloud-config read failure is returned bare, so the user sees the raw file error with no indication it came from --cloud-config. build-iso.go:111 wraps the identical call with "reading cloud config: %w".

Wrap the error the same way build-iso does.

```suggestion
				return fmt.Errorf("reading cloud config: %w", err)
```

pkg/uki/uki.go
itxaka-agent, line 1075: GetUkiCmdline, GetUkiSingleCmdlines (line 1107) and NameFromCmdline (line 1130) all gain a leading baseCmdline parameter. These are exported from pkg/uki, importable by anything outside the repo, so this is a compile break for external callers, for no internal benefit — nothing inside AuroraBoot calls them except Build and the tests.

Thread the computed base through uki.Options instead of changing these exported signatures, or keep the existing three as thin wrappers passing constants.UkiCmdline and add new ...WithBase variants for the SELinux path.

itxaka-agent (changes requested):
> action taken here.

Changes requested — 11 findings inline; two (the `build-uki` flag wiring and the `sdk/utils.OSRelease` fallback path) were independently verified against the pinned toolchain/module before this review was posted.

